PR #8, opened 2017-01-15
Parser for list of parameters

djbe:
Implements these issues:
- https://github.com/AliSoftware/SwiftGen/issues/253
- https://github.com/krzysztofzablocki/Sourcery/issues/14

This parses a list of strings (`Array<String>`) into a dictionary ([String: Any]). You can use it as follows:
```
let template = ...load template...
var context = ...generate context...

context["args"] = Parameters.parse(items: items)
let result = try template.render(context)
```


djbe:
Is this what you guys had in mind? Should I add some sort of method for easily adding this to an (existing) context?

djbe:
invalid keys?

Sources/Parameters.swift
```diff
@@ -0,0 +1,59 @@
...
+		
+		var result = StringDict()
+		for parameter in parameters {
+			result = try parse(item: parameter, result: result)
+		}
+		
+		return result
+	}
+	
+	private static func parse(item: Parameter, result: StringDict) throws -> StringDict {
```
AliSoftware: Maybe rename that private function to `splitAndRepeat` or similar? Making it have the same name as the public one was confusing about its intend on first read

AliSoftware:
I meant whatever use case the `ParameterError.invalidKey(key: String, value: String)` is for.

djbe:
Yeah just saw that. Honestly can't remember when that'd happen. Which means the code needs comments!

AliSoftware:
> Honestly can't remember when that'd happen. Which means the code needs comments!

And tests for those cases you thought about when writing that but forgot since then… making those cases likely to be forgotten again in the future and leading to regressions 😜

Sources/Parameters.swift
```diff
@@ -7,53 +7,67 @@
...
+    let parameters: [Parameter] = try items.map { item in
+      let parts = item.components(separatedBy: "=")
+      guard parts.count == 2 else { throw ParametersError.invalidSyntax(value: item) }
+      return (key: parts[0], value: parts[1])
+    }
+    
+    var result = StringDict()
+    for parameter in parameters {
+      result = try parse(item: parameter, result: result)
+    }
```
AliSoftware: `parameters.reduce(StringDict()) { try Parse(item: $1, résultats: $0) }`?

Sources/Parameters.swift
```diff
@@ -0,0 +1,72 @@
...
+    let parameters: [Parameter] = try items.map { item in
+      let parts = item.components(separatedBy: "=")
+      guard parts.count == 2 else { throw ParametersError.invalidSyntax(value: item) }
+      return (key: parts[0], value: parts[1])
+    }
+    
+    var result = StringDict()
+    for parameter in parameters {
+      result = try parse(parameter: parameter, result: result)
+    }
```
AliSoftware: `parameters.reduce(StringDict()) { try Parse(item: $1, résultats: $0) }`?

djbe: That french autocorrect 😆

Sources/Parameters.swift
```diff
@@ -0,0 +1,72 @@
...
+      let current = result[key] as? StringDict ?? StringDict()
+      let sub = (key: parts.suffix(from: 1).joined(separator: "."), value: parameter.value)
+      result[key] = try parse(parameter: sub, result: current)
+    }
+    
+    return result
+  }
+  
+  // a valid key is not empty and only alphanumerical
+  private static func validate(key: String) -> Bool {
```
AliSoftware: Does this means `1` is a valid key too?

Sources/Parameters.swift
AliSoftware: 😂

Tests/TestSuites/ParametersTests.swift
```diff
@@ -0,0 +1,140 @@
...
+  }
+  
+  func testInvalidSyntax() {
+    do {
+      let items = ["foo:1"]
+      _ = try Parameters.parse(items: items)
+    } catch ParametersError.invalidSyntax {
+      // That's the expected exception we want to happen
+    } catch let error {
+      XCTFail("Unexpected error occured while parsing: \(error)")
```
AliSoftware: Isn't there some `XCTAssertThrows` or similar?

Sources/Parameters.swift
djbe: Right now, yeah. Shouldn't matter that much, say I want to create a (fake) tuple.

Tests/TestSuites/ParametersTests.swift
djbe: No idea, I'll look into it. The structure is just based on the SwiftGenKit (colors) tests

Sources/Parameters.swift
AliSoftware: Ok. Fine by be

Tests/TestSuites/ParametersTests.swift
AliSoftware: That doesn't mean I was right at that time 😂

djbe: And I'm the testing noob 🤣

Sources/Parameters.swift
```diff
@@ -19,8 +19,8 @@ public enum Parameters {
   public static func parse(items: [String]) throws -> StringDict {
     let parameters: [Parameter] = try items.map { item in
       let parts = item.components(separatedBy: "=")
-      if parts.count == 2 {
-        return (key: parts[0], value: parts[1])
+      if parts.count >= 2 {
+        return (key: parts[0], value: parts.dropFirst().joined(separator: "="))
       } else if let part = parts.first, parts.count == 1 && validate(key: part) {
```
AliSoftware: Does that mean that `foo.bar.baz` isn't a valid boolean key? (as `validate(key: "foo.bar.baz")` will return false)?

Sources/Parameters.swift
```diff
@@ -66,9 +66,15 @@ public enum Parameters {
...
   
-  // a valid key is not empty and only alphanumerical
+  // a valid key is not empty and only alphanumerical or dot
   private static func validate(key: String) -> Bool {
     return !key.isEmpty &&
-      key.rangeOfCharacter(from: CharacterSet.alphanumerics.inverted) == nil
+      key.rangeOfCharacter(from: notAlfanumericsAndDot) == nil
   }
+  
+  private static let notAlfanumericsAndDot: CharacterSet = {
```
AliSoftware: s/alfa/alpha 😄

Tests/StencilSwiftKitTests/ParametersTests.swift
```diff
@@ -86,6 +90,7 @@ class ParametersTests: XCTestCase {
   }
   
   func testInvalidKey() {
+    // key may only be alfanumeric or '.'
```
AliSoftware: s/alfa/alpha/

PR #10, opened 2017-01-16
remove join filter

djbe:


djbe:
let's not merge this until after we release swiftgen with great split implementation

AliSoftware:
I've tagged it with `change: breaking` to indicate that it should only be merged when we're ready for a _major_ version bump. Let's finish the Great Split first and release a minor version including that new Great Split reorganisation first, and merge it afterwards for the next major where all templates and documentations would be updated and deprecated tags and filters will be completely removed.

AliSoftware:
Rebased and ready for the next major release

PR #11, opened 2017-01-16
Add map node

djbe:
From the code gen PR: https://github.com/AliSoftware/SwiftGen/pull/188

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,10 @@
 * Added support for Swift Package Manager.  
   [Krzysztof Zabłocki](https://github.com/krzysztofzablocki)
   [#15](https://github.com/SwiftGen/StencilSwiftKit/issues/15)
-* Aded `MapNode` to apply a `map` operator to an array.
-  You can now use `{% map someArray with item set result %}`
+* Added `MapNode` to apply a `map` operator to an array.
+  You can now use `{% map someArray into result using item %}`
   to do the equivalent of the `result = someArray.map { item in … }` Swift code.  
+  [David Jennes](https://github.com/djbe)
   [#11](https://github.com/SwiftGen/StencilSwiftKit/pull/11)
 
 ### Internal Changes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@
 * Added support for Swift Package Manager.  
   [Krzysztof Zabłocki](https://github.com/krzysztofzablocki)
   [#15](https://github.com/SwiftGen/StencilSwiftKit/issues/15)
+* Aded `MapNode` to apply a `map` operator to an array.
+  You can now use `{% map someArray with item set result %}`
+  to do the equivalent of the `result = someArray.map { item in … }` Swift code.  
+  [#11](https://github.com/SwiftGen/StencilSwiftKit/pull/11)
 
 ### Internal Changes
 
--- a/README.md
+++ b/README.md
@@ -19,6 +19,9 @@ _TODO: [Write more extension Documentation](https://github.com/SwiftGen/StencilS
 * `SetNode`
   * `{% set <Name> %}…{% endset %}`
   * Renders the nodes inside this block immediately, and stores the result in the `<Name`>  variable of the current context.
+* `MapNode`
+  * `{% map <Variable> into <Name> using <ItemName> %}…{% endmap %}`
+  * Apply a `map` operator to an array, and store the result into a new array variable `<Name>` in the current context.
 
 ## Filters
 
--- a/README.md
+++ b/README.md
@@ -22,6 +22,7 @@ _TODO: [Write more extension Documentation](https://github.com/SwiftGen/StencilS
 * `MapNode`
   * `{% map <Variable> into <Name> using <ItemName> %}…{% endmap %}`
   * Apply a `map` operator to an array, and store the result into a new array variable `<Name>` in the current context.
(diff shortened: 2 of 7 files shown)


Sources/MapNode.swift
```diff
@@ -0,0 +1,65 @@
+//
+// StencilSwiftKit
+// Copyright (c) 2017 Olivier Halligon
+// Created by Peter Livesey.
+// MIT Licence
+//
+
+import Stencil
+
+public class MapNode: NodeType {
```
AliSoftware: I wonder if the name of the variable to store the result of the map should be part of the Map node?

I mean, that node syntax reads strange to be and doesn't seem that natural to use (`{% map array with element set varname %}`, I always wonder what `element` and `varname` are, which is which, etc). And the node doesn't produce any output as a result, storing the result in the context instead, that seems convoluted at first read.

That would be awesome if it would be possible to just use `{% map array with x %}Item {{maploop.counter}} = {{x}}.{% endmap %}` and that would just generate / output… an array. That the user can then just use as-is (let it render), or store in a variable using the `SetNode`.

I know that's not really possible now, as `render` outputs a `String` (which is kinda logical), and making it a Filter instead of a Node won't make sense either because we need to let us pass a whole Stencil tree as the "closure" of that "map" — so `{{ array|map:what_to_put_here }}` wouldn't cut it (except… if `what_to_put_here` was a `CallableNode` from a `MacroNode`, heh!) — but I'm wondering if we couldn't find a more elegant solution that this `{% map array with item set result %}`?

\cc @kylef thoughts?

AliSoftware:
I wonder if the name of the variable to store the result of the map should be part of the Map node?

I mean, that node syntax reads strange to be and doesn't seem that natural to use (`{% map array with element set varname %}`, I always wonder what `element` and `varname` are, which is which, etc). And the node doesn't produce any output as a result, storing the result in the context instead, that seems convoluted at first read.

That would be awesome if it would be possible to just use `{% map array with x %}Item {{maploop.counter}} = {{x}}.{% endmap %}` and that would just generate / output… an array. That the user can then just use as-is (let it render), or store in a variable using the `SetNode`.

I know that's not really possible now, as `render` outputs a `String` (which is kinda logical), and making it a Filter instead of a Node won't make sense either because we need to let us pass a whole Stencil tree as the "closure" of that "map" — so `{{ array|map:what_to_put_here }}` wouldn't cut it (except… if `what_to_put_here` was a `CallableNode` from a `MacroNode`, heh!) — but I'm wondering if we couldn't find a more elegant solution that this `{% map array with item set result %}`?

\cc @kylef thoughts?

djbe:
That's one of the reasons why I ended up creating a modified set that accepts a 2nd argument instead of a body, to avoid losing type information.

What we can do is simplify the map syntax of map to:
```{% map array into result %}{{maploop.item|lowercase}}{% endmap %}```
Not necessarily removing the option of naming the loop variable, but making it optional. 

When the loop item name isn't specified we can either:
- provide a default name such as `item`
- only store the value inside `maploop.item`
And still allow the user to provide a name using the `with myVarName` arguments.

**Comment by AliSoftware** [comment]
I like that `{% map array into result [using var] %}` much better than `{% map array with var set result %}` indeed. More understandable imho. And if the optional `using var` part isn't provided, `maploop.item` seems the best option to me (to avoid potential collision with any `item` variable that could have been set outside of the map and better scope everything)

**Comment by djbe** [comment]
Cool! I'll make those changes then.

File: CHANGELOG.md [hunk 2]
```diff
@@ -8,6 +8,11 @@
 
 ### New Features
 
+* Aded `MapNode` to apply a `map` operator to an array.
```
AliSoftware: s/Aded/Added/

Tests/TestSuites/MapNodeTests.swift
```diff
@@ -0,0 +1,31 @@
...
+    "items1": ["one", "two", "three"],
+    "items2": ["hello", "world", "everyone"]
+  ]
+  
+  func testBasic() {
+    let template = SwiftTemplate(templateString: Fixtures.string(for: "map-basic.stencil"), environment: stencilSwiftEnvironment())
+    let result = try! template.render(MapNodeTests.context)
+    
+    let expected = Fixtures.string(for: "map-basic.out")
+    XCTDiffStrings(result, expected)
```
AliSoftware: I'm wondering if testing Stencil Nodes using templates is a good thing.
When comes the time when we decide to move the MapNode to Stencil itself, Stencil won't have our fixtures as templates to test anyway.

The way Stencil tests its nodes is by invoking the Node constructor, then its render method, and test the result, without needing a template fixture for node unit tests. [See here](https://github.com/kylef/Stencil/blob/master/Tests/StencilTests/ForNodeSpec.swift#L14-L16). Maybe we should test at the same level too.

djbe: I'll take a look at the Stencil tests.

I'll have to try and convert the current tests for this (and other nodes) to comparable tests using constructors and such.

README.md
```diff
@@ -19,6 +19,9 @@ _TODO: [Write more extension Documentation](https://github.com/SwiftGen/StencilS
 * `SetNode`
   * `{% set <Name> %}…{% endset %}`
   * Renders the nodes inside this block immediately, and stores the result in the `<Name`>  variable of the current context.
+* `MapNode`
+  * `{% map <Variable> into <Name> using <ItemName> %}…{% endmap %}`
+  * Apply a `map` operator to an array, and store the result into a new array variable `<Name>` in the current context.
```
AliSoftware: + * Inside the map loop, a `maploop` special variable is available (akin to the `forloop` variable in `for` nodes). It exposes `maploop.count`, `maploop.first`, `maploop.last` and `maploop.item`.

PR #15, opened 2017-01-27
feat: add support for Swift Package Manager

krzysztofzablocki:
Ads suport for Swift Package Manager


Sources/SwiftTemplate.swift
```diff
@@ -3,9 +3,16 @@
 // Copyright (c) 2017 SwiftGen
 // MIT Licence
 //
-
+import Foundation
 import Stencil
 
+#if os(Linux)
+#if swift(>=3.1)
```
AliSoftware: Indeed `#if swift(<3.1)` isn't a valid syntax… But I think `#if ! swift(>=3.1)` is. This would avoid an empty `#if` body and the main logic in the `#else`

You could even probably just write `#if os(Linux) && ! swift(>=3.1)`

djbe: See the spm branch I made, is esentially this + tests

AliSoftware: Is it a branch that is based on @krzysztofzablocki 's `feature/spm-support` branch of this PR or a separate branch from `develop`? (Sorry I'm in the train, not enough connection for git fetch to work reliably to check myself 😄)

@djbe This branch is on the repo (and not a fork), and you both have ( @krzysztofzablocki + you) push access on this repo, so you should be able to add commits to this branch instead of creating a separate one, any reason not to here?

djbe: only because I was playing around and wanted to let him compare. I'll delete mine and push the canges here, gimme a moment while I rebase the rest of the branches (so many branches :P )

krzysztofzablocki: Might be worth adding `swift build`/test to avoid regression

djbe: Dunno, Rakefile (Xcode project) and swift package test the same thing. Testing both would just double the length of the CI tests.

krzysztofzablocki: I think its important because dependency bindings between podspec and spm are different, you might end up in situation when we update podspec dependency and forget SPM one and swift test / build would prevent it

djbe: Can you add it to the rakefile? `swift build && swift test`
My `rake test` is broken for some reason (I really really really hate ruby version shizzle)

krzysztofzablocki:
not afaik

krzysztofzablocki:
can we merge this?

PR #16, opened 2017-01-27
Refactor stencil swift extensions registration for easier use with an existing `Extension`.

djbe:


djbe:
will do

djbe:
Do note that I can't merge, only you can.

PR #18, opened 2017-02-04
JavaScriptCore support

ilyapuchka:
This is an implementation of JavaScriptCore support, mentioned in https://github.com/krzysztofzablocki/Sourcery/issues/135 that adds custom filters and tags.


ilyapuchka:
@AliSoftware @djbe can some one review this?

djbe:
Will check it out later in the weekend if I find some time. I've already seen the initial commits a while ago and those looked ok (from a quick glance).

The only thing I'm really really wary of is, what is the performance implication of this? I mean, it's JavaScript, that stuff by definition is slow as molasses 😄 (exaggerating, but you get my point?)

AliSoftware:
Sorry for the delay @ilyapuchka , we're quite focused on finishing the Great Split (waiting for Kyle to release Stencil 0.8, hopefully tonight or early tomorrow) so we can finally (!) merge & freeze & tag & push everything, so we can have a clean setup again and release that long-awaited stable 4.2 version at last. That's why that PR hasn't been our priority 😉

Then we'll focus on repairing the CI and make it possible to ensure PRs don't break anything, so we can work on PRs serenely again, because our Travis setup is quite unstable and not trustworthy lately 😅

ilyapuchka:
if it's ok for react native, I guess it will be ok for code generator tasks performance wise =)

Sources/JSEnvironment.swift
```diff
@@ -0,0 +1,332 @@
...
+class JSTokenParser: NSObject, JSExportableTokenParser {
+    let parser: TokenParser
+    init(_ parser: TokenParser) {
+        self.parser = parser
+    }
+
+    func parse() -> [JSNode] {
+        return bridgingError {
+            try parser.parse().map(JSNode.init)
+            } ?? []
```
djbe: Indentation

Tests/StencilSwiftKitTests/JSFiltersTests.swift
```diff
@@ -0,0 +1,49 @@
...
+// Copyright (c) 2017 SwiftGen
+// MIT Licence
+//
+
+import XCTest
+import StencilSwiftKit
+
+class JSFiltersTests: XCTestCase {
+  
+  func testJSFilter() {
```
djbe: Please rewrite the tests to be similar to the other node/filter tests. This means to write tests with a few parts:
- given a list of tokens, parses a correct structure of nodes/filters and such
- given a structure of nodes and a context, renders the correct output
- some tests to verify that parsing/rendering correctly fails for invalid data.

ilyapuchka:
@djbe My vision exactly what you state as a second variant. Custom JS filters should be loaded by scanning some folder that user provides (with a command line option most likely). But it's up to the tool (consumer of this framework) to define how exactly it will be loaded. That's why it's not included in this PR as I believe its out of the scope of this framework. I will go through your comments latter. Thanks for the review!

Sources/Environment.swift
```diff
@@ -25,9 +25,9 @@ public extension Extension {
   }
 }
 
-public func stencilSwiftEnvironment() -> Environment {
+public func stencilSwiftEnvironment(_ configure: (Extension) -> () = { _ in }) -> Environment {
```
ilyapuchka: `extensions` is defined as immutable:

```swift
public struct Environment {
  public let templateClass: Template.Type
  public let extensions: [Extension]
```

Sources/JSEnvironment.swift
```diff
@@ -0,0 +1,332 @@
+//
```
ilyapuchka: Moving JS wrappers to separate files (wrapper per file) will be ok?

djbe: To begin with, yeah, after that we'll see about moving it to a submodule. I can imagine some people using StencilSwiftKit on platforms where there's no JavascriptCore support.

Sources/Environment.swift
djbe: Allright, so if it's immutable, modify this function to accept an existing list of extensions (default to empty array). Adding this closure to configure an extension just seems unnecessary and awkward.

ilyapuchka:
@djbe I've addressed your comments and added more tests, please review my changes

ilyapuchka:
@AliSoftware @djbe any more thoughts about this? 😉

djbe:
Sorry, the last few weeks have been really busy for me. I'll take a look at your changes this weekend.

Sources/JSExtension.swift
```diff
@@ -0,0 +1,338 @@
+//
```
djbe: This is still one big file, please split this up as mentioned before, one file for each logical block. For example, move the JS wrappers to another file.

Sources/JSExtension.swift
```diff
@@ -0,0 +1,338 @@
...
+    /// ```
+    /// function greet(parser, token) {
+    ///  // use parser and token to parse tag
+    ///
+    ///  this.render = function(context) {
+    ///    // return rendered content as a string
+    ///  }
+    /// }
+    /// ```
+    public func registerTag(_ name: String, script code: String) {
```
djbe: Are all the wrappers below (such as JSTokenParser, JSToken, ...) needed for only this call? If so, we might want to re-think it. I'm not a fan of repeating and creating copies of everything in Stencil, just to support this. What if stencil adds new types, or changes how they work?

Maybe a more limited version would be preferable. For example, define a way that a javascript function/object can define a list of parameters and a render function. And then unify this with the "simple tag" you have above.

Tests/StencilSwiftKitTests/JSExtensionTests.swift
```diff
@@ -0,0 +1,151 @@
...
+// Copyright (c) 2017 SwiftGen
+// MIT Licence
+//
+
+import XCTest
+import StencilSwiftKit
+
+class JSExtensionTests: XCTestCase {
+
+  func template(_ templateString: String,
```
djbe: You've added more tests, which is good, but you're still missing (mentioned before):
- given a list of tokens, parses a correct structure of nodes/filters and such
- given a structure of nodes and a context, renders the correct output

You're directly going from stencil template string to a render. We prefer to test this in 2 fases: parsing, and then rendering.

ilyapuchka:
@djbe this code is already extracted into a framework - https://github.com/ilyapuchka/StencilJS so you can use it instead of merging this pr if you are interested in this feature. I don't think we will need this for Sourcery though, what was initial intention for this pr, as we already have EJS and Swift templates  support which both look much more promising than Stencil.

djbe:
Oh, ok then.

As you're no longer pushing for this feature, and unless anyone wants it, I'll consider this closed. We can always add StencilJS as a dependency at some point.

PR #19, opened 2017-02-15
small fix to match stencil context type

djbe:


AliSoftware:
Hello Travis my old friend?! 😒

Sources/Context.swift
```diff
@@ -8,7 +8,7 @@
 
 import Foundation
 
-public func enrich(context: [AnyHashable: Any], parameters: [String]) throws -> [AnyHashable: Any] {
+public func enrich(context: [String: Any], parameters: [String]) throws -> [String: Any] {
   var context = context
   
   context["env"] = ProcessInfo().environment
```
AliSoftware: Make "env" and "param" constants instead of using magic strings

Sources/Context.swift
```diff
@@ -8,7 +8,7 @@
 
 import Foundation
 
-public func enrich(context: [AnyHashable: Any], parameters: [String]) throws -> [AnyHashable: Any] {
+public func enrich(context: [String: Any], parameters: [String]) throws -> [String: Any] {
```
AliSoftware: add `env` as a parameter to `enrich`, with default value `ProcessInfo().environment`, both for testability and in case other apps/clients of the framework want to alter the env before enriching (to hide some sensible env vars for example)

Tests/StencilSwiftKitTests/ContextTests.swift
```diff
@@ -16,9 +16,9 @@ class ContextTests: XCTestCase {
     let result = try! enrich(context: context, parameters: [])
     XCTAssertEqual(result.count, 2, "2 items have been added")
     
-    guard let env = result["env"] as? [AnyHashable: Any] else { XCTFail("`env` should be a dictionary"); return }
+    guard let env = result["env"] as? [String: Any] else { XCTFail("`env` should be a dictionary"); return }
```
AliSoftware: use the new `StencilContext.environment` constant there

Tests/StencilSwiftKitTests/ContextTests.swift
```diff
@@ -16,9 +16,9 @@ class ContextTests: XCTestCase {
...
     XCTAssertNotNil(env["PATH"] as? String)
-    guard let params = result["param"] as? [AnyHashable: Any] else { XCTFail("`param` should be a dictionary"); return }
+    guard let params = result["param"] as? [String: Any] else { XCTFail("`param` should be a dictionary"); return }
```
AliSoftware: use the new `StencilContext.parameters` constant there

Sources/Context.swift
```diff
@@ -8,11 +8,23 @@
...
+}
+
+/**
+ Enriches a stencil context with parsed parameters and environment variables
+ 
+ - Parameter context: The stencil context
+ - Parameter parameters: List of strings, will be parsed using the `Parameters.parse(items:)` method
+ - Parameter environment: Environment variables, defaults to `ProcessInfo().environment`
+ */
+public func enrich(context: [String: Any], parameters: [String], environment: [String: String] = ProcessInfo().environment) throws -> [String: Any] {
```
AliSoftware: Now that we have an enum, and because I'm not a fan of top-level functions, maybe it would be nice to move the `enrich` func as a `static func` inside `enum StencilContext`?

Tests/StencilSwiftKitTests/ContextTests.swift
```diff
@@ -13,38 +13,38 @@ class ContextTests: XCTestCase {
   func testEmpty() {
     let context = [String: Any]()
     
-    let result = try! enrich(context: context, parameters: [])
+    let result = try! StencilContext.enrich(context: context, parameters: [])
```
AliSoftware: Shouldn't we explicitly pass a `["PATH": "foo:bar:baz"]` value there for the `environment:` parameter, so that the unit test is deterministic and not dependent on the shell environment used by the test process? (even if `PATH` is really likely to be present in any shell environment, at least there it's explicit)?

PR #20, opened 2017-02-15
Configure Circle CI

djbe:
Note: Once this is merged, don't forget to:

* [ ] Adapt the GitHub settings of this repo to make circle the mandatory status check instead of travis


AliSoftware:
We should create the Rakefle for each repo before starting to adjust the circle.yml files: it would be more readable for `circle.yml` to invoke  rake test` than the whole `xcodebuild` command

djbe:
Sure, but we should add versions depending on environment. If CI, output logs to `$CIRCLE_ARTIFACTS` and `$CIRCLE_TEST_REPORTS`. Also each rake task should be separate, so we can eventually parallelize stuff.

Rakefile
```diff
@@ -1,25 +1,42 @@
...
+task :xcode_test => :xcode_build do
+  xcpretty("xcodebuild -workspace StencilSwiftKit.xcworkspace -scheme Tests test-without-building", "xcode_test")
 end
 
-desc 'Run Unit Tests'
-task :test => :build_for_testing do
-  sh "swift test"
-  xcpretty "xcodebuild -workspace StencilSwiftKit.xcworkspace -scheme Tests test-without-building"
+desc 'Run SPM Unit Tests'
+task :spm_test => :spm_build do
```
AliSoftware: Tip: you can use take namespaces to group relative tasks together.

A `namespace 'spm' do … end` and a `namespace 'xcodebuild' do … end` would be nicer to read, and then you can name your tasks just `build` and `test` and refer to them as `rake spm:test` etc.

djbe:
Only missing thing is a review from Danielle and we're good to go.

circle.yml
```diff
@@ -0,0 +1,17 @@
+machine:
+  xcode:
+    version: 8.2
+
+dependencies:
+  pre:
+    - gem install xcpretty --no-rdoc --no-ri --no-document --quiet
+    - if [[ $CIRCLE_BRANCH == master ]]; then pod setup; fi
+  cache_directories:
+    -  "~/.cocoapods/repos/master"
```
AliSoftware: @djbe Well I think we still need to change that `pod setup` + `cache_directories` strategy to use what @dantoml told us about in Slack, to improve that part before consider merging.

djbe: That's what I meant with input from @dantoml 😄

Rakefile
```diff
@@ -1,25 +1,48 @@
...
-  xcpretty "xcodebuild -workspace StencilSwiftKit.xcworkspace -scheme Tests test-without-building"
+namespace :spm do
+  desc 'Build using SPM'
+  task :build do
+    plain("swift build", "spm_build")
+  end
+
+  desc 'Run SPM Unit Tests'
+  task :test => :build do
+    plain("swift test", "spm_build")
```
AliSoftware: Shouldn't the `name` param be `spm_test` there instead of `spm_build` (especially to avoid the risk of overriding the log output from build)?

djbe: I tried using a ruby trick to get the calling method's name, but keep getting `block (2 levels) in <top (required)>`, so gave up on that.

AliSoftware: You could do that by capturing the 2st argument passed to the block, which is the rake task object itself, which as a `.name` property:

```
namespace :log do
  task :foo do |t|
    puts "Name: #{t.name}" // t.name is indeed "log:foo"
  end
end
```

Rakefile
```diff
@@ -42,9 +42,16 @@ namespace :xcode do
...
+namespace :lint do
+  desc 'Lint the Pod'
+  task :pod do |task|
+    plain("pod lib lint StencilSwiftKit.podspec --quick", task)
+  end
+  
+  desc 'Lint the code'
+  task :code do |task|
+    plain("PROJECT_DIR=. ./Scripts/swiftlint-code.sh", task)
+  end
```
AliSoftware: Just a thought: I'm not sure about the readability/logic of grouping those two commands, still undecided.

Because "linting a pod" and "linting the code" seems two different kinds of lints to me. One is checking that the podspec "compiles" (the podspec file doesn't have any syntax error, and that a project with that pod would compile), the other is checking for code style issues.

Maybe it would be more logical to put everything related to SwiftLint in the Rakefile anyway (I mean instead of having separate `.sh` scripts to install swiftlint and lint the code — which were there at first because we didn't have any Rakefile or because we found this install_swiftlint script as is) and have `rake swiftlint:install`, `rake swiftlint:sourcecode` and `rake swiftlint:tests` tasks? (and then concerning the `pod lib lint`, not sure it's worth having a dedicated rake task?)

StencilSwiftKit.xcworkspace/xcshareddata/StencilSwiftKit.xcscmblueprint
```diff
@@ -0,0 +1,30 @@
+{
+  "DVTSourceControlWorkspaceBlueprintPrimaryRemoteRepositoryKey" : "804E0A642F9B0288E24EC6843BD9BD71D8BD7EA7",
```
AliSoftware: I never know if this file should be pushed on git, I see it being ignored by a lot of .gitignored on OSS projects and I generally gitignore it too, I think because otherwise it gets updated almost on every commit… I believe it's only useful for Xcode to have metadata about the repo, but who uses Xcode's integrated SCM tool anyway

Tests/StencilSwiftKitTests/CallNodeTests.swift
```diff
@@ -11,64 +11,64 @@ import XCTest
...
-      
+
       let parser = TokenParser(tokens: tokens, environment: stencilSwiftEnvironment())
       XCTAssertThrowsError(try parser.parse())
     }
   }
-  
-  func testRender() {
+
+  func testRender() throws {
```
AliSoftware: I didn't even know XCTest would still recognize functions as tests when they are marked with `throws`, I thought only `func test<Whatever>()` functions with this exact signature type would be recognized! (wait, that test actually _does_ get executed, right?)

Tests/StencilSwiftKitTests/TestsHelper.swift
```diff
@@ -42,7 +45,12 @@ func diff(_ result: String, _ expected: String) -> String? {
     }
     let lhsNum = addLineNumbers(slice(lhsLines, 4)).joined(separator: "\n")
     let rhsNum = addLineNumbers(slice(rhsLines, 4)).joined(separator: "\n")
-    return "\(msgColor)Mismatch at line \(badLineIdx)\(reset)\n>>>>>> result\n\(lhsNum)\n======\n\(rhsNum)\n<<<<<< expected"
+    return "\(msgColor)Mismatch at line \(badLineIdx)\(reset)\n" +
+      ">>>>>> result\n" +
+      "\(lhsNum)\n" +
+      "======\n" +
+      "\(rhsNum)\n" +
+      "<<<<<< expected"
```
AliSoftware: I think an array of strings that gets `join`'d is more efficient that successive string concatenation… and compiles faster (because `+` is so overloaded) too

circle.yml
```diff
@@ -5,13 +5,15 @@ machine:
 dependencies:
   pre:
     - gem install xcpretty --no-rdoc --no-ri --no-document --quiet
+    - Scripts/install_swiftlint.sh
     - if [[ $CIRCLE_BRANCH == master ]]; then pod setup; fi
```
AliSoftware: This needs to be converted to the "download_from_s3" trick

circle.yml
```diff
@@ -5,13 +5,15 @@ machine:
...
     -  "~/.cocoapods/repos/master"
 
 test:
   override:
+    - rake lint:code
     - rake xcode:test
     - rake spm:test
   post:
-    - if [[ $CIRCLE_BRANCH == master ]]; then rake lint; fi
+    - if [[ $CIRCLE_BRANCH == master ]]; then rake lint:pod; fi
```
AliSoftware: Other alternative: `[[ $CIRCLE_BRANCH == master ]] && rake lint:post` (which would actually be `[[ $CIRCLE_BRANCH == master ]] && pod lib lint --quick` if we take the pod lint out of the Rakefile).
I have no real preference, just saying in case you prefer `&&` over `if; then; fi`

StencilSwiftKit.xcworkspace/xcshareddata/StencilSwiftKit.xcscmblueprint
djbe: Huh, I thought I deleted that. Must be in another branch

circle.yml
djbe: Jup, added it while you were typing 😆

Tests/StencilSwiftKitTests/CallNodeTests.swift
djbe: I saw it as a little know trick recently related to XCTest, maybe one of Natasha's newsletters or something? Can't remember.

Anyway, the test still gets executed, during testing it just has a different name (`AndReturnError:` gets appended): `[Tests.CallNodeTests testRenderAndReturnError:]`

AliSoftware: 👍

Tests/StencilSwiftKitTests/TestsHelper.swift
```diff
@@ -42,7 +45,14 @@ func diff(_ result: String, _ expected: String) -> String? {
     }
     let lhsNum = addLineNumbers(slice(lhsLines, 4)).joined(separator: "\n")
     let rhsNum = addLineNumbers(slice(rhsLines, 4)).joined(separator: "\n")
-    return "\(msgColor)Mismatch at line \(badLineIdx)\(reset)\n>>>>>> result\n\(lhsNum)\n======\n\(rhsNum)\n<<<<<< expected"
+    return [
+      "\(msgColor)Mismatch at line \(badLineIdx)\(reset)\n",
```
AliSoftware: Extra `\n` (which should be added by the `join`: is it intentional (so that it ends up with two \n for this line and thus a blank line in the result) or not?

djbe: No, not intentional ☹️

djbe:
How do we merge this PR? The github configuration'll need to be changed to require circle tests instead of travis before we can merge it.

AliSoftware:
Done :)

PR #24, opened 2017-02-22
Added StringFilters.camelToSnakeCase filter

ggrell:
Registered for Stencil Swift extensions
Added test coverage
Updated README.md


Sources/Filters.swift
```diff
@@ -89,6 +89,11 @@ struct StringFilters {
     }
   }
 
+  static func camelToSnakeCase(_ value: Any?) throws -> Any? {
+    guard let string = value as? String else { throw FilterError.invalidInputType }
+    return try snakecase(string).lowercased()
```
ggrell: Hmm, don't think I've ever seen snake case with caps. The scenario I'm trying to get to is auto-generating json_field_names from camelCase, which tend to be all lower case.

Should I break it out into two different ones? camelToSnakeCase and camelToLowerSnakeCase? Or does Stencil already provides a way to combine something like this: {{ "someTextHere"|camelToSnakeCase|lowercase }}?

djbe: I haven't seen it either, I was just wondering what the correct / default behaviour should be. Looking at some libraries in other languages, the default seems to be lowercase, so let's do that 👍 

You could always add a parameter to the filter to disable the lowercasing (see http://stencil.fuller.li/en/latest/builtins.html#built-in-filters).

ggrell: Unfortunately the Wikipedia page doesn't give any definitive rules for upper snake case. For instance, as the StringFilter.snakecase() function is currently written, URLChooser becomes URL_Chooser, not Url_Chooser. I couldn't really find anything that said that'd be right or wrong, so I think for now I'll just stick with that implementation. The filter itself will take one optional boolean parameter for lowercasing, defaults to true.

README.md
```diff
@@ -31,7 +31,8 @@ _TODO: [Write more extension Documentation](https://github.com/SwiftGen/StencilS
 * `swiftIdentifier`: Transforms an arbitrary string into a valid Swift identifier (using only valid characters for a Swift identifier as defined in the Swift language reference)
 * `join`: Deprecated. Will be removed now that the same filter exists in Stencil proper.
 * `lowerFirstWord`
-* `snakeToCamelCase` / `snakeToCamelCaseNoPrefix`
+* `snakeToCamelCase` / `snakeToCamelCaseNoPrefix` /
```
djbe: Maybe remove that trailing `/`

StencilSwiftKit.xcodeproj/xcshareddata/xcschemes/Tests.xcscheme
```diff
@@ -8,7 +8,7 @@
       <BuildActionEntries>
          <BuildActionEntry
             buildForTesting = "YES"
-            buildForRunning = "NO"
+            buildForRunning = "YES"
```
djbe: Is there a reason this was changed? I don't think it does much.

Sources/Filters.swift
```diff
@@ -89,6 +89,34 @@ struct StringFilters {
...
+  /// Converts camelCase to snake_case. Takes an optional Bool argument for making the string lower case,
+  /// which defaults to true
+  ///
+  /// - parameter value: the value to be processed
+  /// - parameter arguments: the arguments to the function, expecting zero or one argument
+  /// - returns: the snake case string
+  /// - throws: FilterError.invalidInputType if the value parameter isn't a string
+  static func camelToSnakeCase(_ value: Any?, arguments: [Any?]) throws -> Any? {
+    var toLower = true
+    if arguments.count ==  1 {
```
djbe: It might be useful to refactor out this bit (to parse a boolean parameter). That way you can also simplify your tests.

Tests/StencilSwiftKitTests/StringFiltersTests.swift
```diff
@@ -133,6 +133,84 @@ class StringFiltersTests: XCTestCase {
...
+            "a__b__c": "a__b__c",
+            "__y_z!": "__y_z!",
+            "PLEASESTOPSCREAMING": "pleasestopscreaming",
+            "PLEASESTOPSCREAMING!": "pleasestopscreaming!",
+            "PLEASE_STOP_SCREAMING": "please_stop_screaming",
+            "PLEASE_STOP_SCREAMING!": "please_stop_screaming!"
+    ]
+
+    for (input, expected) in expectations {
+      let trueArgResult = try! StringFilters.camelToSnakeCase(input, arguments: ["true"]) as? String
```
djbe: As mentioned above, it might be better to separate the boolean parsing logic into it's own method, that can then be tested separately.

README.md
ggrell: Yup will do

StencilSwiftKit.xcodeproj/xcshareddata/xcschemes/Tests.xcscheme
ggrell: Ah, probably AppCode made then change since I was running unit tests. Will revert.

Sources/Filters.swift
```diff
@@ -89,6 +89,44 @@ struct StringFilters {
...
+    return snakeCase
+  }
+
+  /// Parses filter arguments for a boolean value, where true can by any one of: "true", "yes", "1", and
+  /// false can be any one of: "false", "no", "0".
+  /// - parameter arguments: an array of argument values, may be empty
+  /// - parameter index: the index in the arguments array
+  /// - returns: true or false if a value was parsed, or nil if it wasn't able to
+  static func parseBool(from arguments: [Any?], index: Int) -> Bool? {
+    guard index + 1 <= arguments.count else {
```
djbe: `index < arguments.count`

Sources/Filters.swift
```diff
@@ -89,6 +89,44 @@ struct StringFilters {
...
+    }
+
+    let boolArg = (arguments[index] as? String ?? "").lowercased()
+    switch boolArg {
+    case "false", "no", "0":
+      return false
+    case "true", "yes", "1":
+      return true
+    default:
+      return nil
```
djbe: Should the default here be `nil` or throwing an exception?

Tests/StencilSwiftKitTests/StringFiltersTests.swift
```diff
@@ -133,6 +133,122 @@ class StringFiltersTests: XCTestCase {
     }
   }
 
+  func testParseBool_WithTrueString() throws {
```
djbe: Might be better to move all these parse tests to a separate XCTestCase.

Sources/Filters.swift
ggrell: Hmm. I'll refactor. If it's an optional argument (which I tested that Stencil allows for it), it should return nil, otherwise it should throw.

djbe: Sounds good!

Sources/Filters.swift
```diff
@@ -123,7 +129,11 @@ struct StringFilters {
     case "true", "yes", "1":
       return true
     default:
-      return nil
+      if required {
```
djbe: What I meant with my throwing behaviour was more along the lines of, if I pass the parameter `someInvalidStuff` instead of `true` or `false`, that should be considered and thus throw.

I'd change the guard to:
```swift
guard index < arguments.count, let boolArg = arguments[index] as? String else {
  ... required check...
```
In the switch, you then don't check for `required`, the default case always throws.

Tests/StencilSwiftKitTests/StringFiltersTests.swift
```diff
@@ -133,6 +133,77 @@ class StringFiltersTests: XCTestCase {
     }
   }
 
+  func testCamelToSnakeCase_WithNoArgsDefaultsToTrue() {
```
djbe: You can remove this case, it's covered by the parseBool tests and the 2 cases below this one.

Sources/Filters.swift
ggrell: Yup, good eye!

Tests/StencilSwiftKitTests/StringFiltersTests.swift
ggrell: This test covers the ```?? true``` flow in the line:
```let toLower = try parseBool(from: arguments, index: 0, required: false) ?? true``` because there's no other test for the no arguments case. The other two camelToSnakeCase tests pass arguments.

Comment (djbe) [inline Tests/StencilSwiftKitTests/StringFiltersTests.swift]: True, good call.

File: Sources/Filters.swift [hunk 11]
```diff
@@ -89,6 +89,50 @@ struct StringFilters {
...
+  /// - parameter arguments: the arguments to the function, expecting zero or one argument
+  /// - returns: the snake case string
+  /// - throws: FilterError.invalidInputType if the value parameter isn't a string
+  static func camelToSnakeCase(_ value: Any?, arguments: [Any?]) throws -> Any? {
+    let toLower = try parseBool(from: arguments, index: 0, required: false) ?? true
+    guard let string = value as? String else { throw FilterError.invalidInputType }
+
+    let snakeCase = try snakecase(string)
+    if toLower {
+      return snakeCase.lowercased()
```
AliSoftware: There's already a `lowercase` filter provided by Stencil, so you could already do `foo|camelToSnakeCase|lowercase` instead of `foo|camelToSnakeCase(true)`.

I think it would be better to keep the lowercasing separate, i.e. let the caller filter thru `|lowercase` in the template, instead of making that logic in code (and with the risk of repeating lowercasing logic in Swift code).

Besides, it would allow you to simplify this whole PR by not needing this `parseBool` anymore (I'm surprised that `Stencil` doesn't provide a way to parse a boolean argument the same way for all filters already btw?).

And I think it would be more understandable to be explicit. Sure you'll lose the "automatic lowercase by default" behavior, that would be unfortunate, but the code would be much simpler (and the risk of regression and inconsistency in the bool argument parsing logic lowered too)

Tests/StencilSwiftKitTests/StringFiltersTests.swift
```diff
@@ -133,6 +133,77 @@ class StringFiltersTests: XCTestCase {
...
+
+    for (input, expected) in expectations {
+      let trueArgResult = try StringFilters.camelToSnakeCase(input, arguments: ["true"]) as? String
+      XCTAssertEqual(trueArgResult, expected)
+    }
+  }
+
+  func testCamelToSnakeCase_WithFalse() throws {
+    let expectations = [
+            "string": "string",
```
AliSoftware: A little nitpicking, but could you ensure the indentation you use is consistent with the rest of the project? Thx
(Here even in your own code you have different indentation levels)

Tests/StencilSwiftKitTests/XCTest+Assertions.swift
```diff
@@ -0,0 +1,12 @@
+//
+// StencilSwiftKit
+// Copyright (c) 2017 SwiftGen
+// MIT Licence
+//
+
+import XCTest
+
+public func XCTAssertThrows(_ expression: @autoclosure () throws -> Any?, _ message: @autoclosure () -> String = "",
```
AliSoftware: Could you instead move that into the `TestsHelper.swift` existing file?

(tbh I was sure we already implemented a similar helper somewhere but maybe it was in another PR?!)

Sources/Filters.swift
djbe: We might need the parseBool logic for other filters we want to refactor (snakeToCamelCaseNoPrefix), but it might then be better to leave that for another PR.

Tests/StencilSwiftKitTests/XCTest+Assertions.swift
djbe: I remember discussing this, but for now we use `XCTAssertThrowsError` when we want / expect the expression to fail with an exception.

Sources/Filters.swift
ggrell: Currently `foo|camelToSnakeCase|lowerCase` and `foo|camelToSnakeCase` are equivalent which I expect that 99% of people will want. You would only need to add a parameter if you wanted snake case with first upper characters, which IMHO will probably be pretty rare. That's why I'd rather keep it as is.

Tests/StencilSwiftKitTests/XCTest+Assertions.swift
ggrell: Wow I don't even remember why I ended up writing it. I'll remove it and just use `XCTAssertThrowsError` instead.

ggrell:
Code is ready to be merged, thanks!

StencilSwiftKit.xcodeproj/xcshareddata/xcschemes/Tests.xcscheme
AliSoftware: Seems to still be there?

Sources/Filters.swift
```diff
@@ -89,6 +89,50 @@ struct StringFilters {
...
+
+  /// Parses filter arguments for a boolean value, where true can by any one of: "true", "yes", "1", and
+  /// false can be any one of: "false", "no", "0". If optional is true it means that the argument on the filter is
+  /// optional and it's not an error condition if the argument is missing or not the right type
+  /// - parameter arguments: an array of argument values, may be empty
+  /// - parameter index: the index in the arguments array
+  /// - parameter required: If true, the argument is required and function throws if missing.
+  ///                       If false, returns nil on missing args.
+  /// - returns: true or false if a value was parsed, or nil if it wasn't able to
+  static func parseBool(from arguments: [Any?], index: Int, required: Bool = true) throws -> Bool? {
```
AliSoftware: As we're gonna keep this `parseBool` method, there's no reason for it to be only limited to string filters so no reason to scope it in the `StringFilters` type. May be worth creating a dedicated type for filter helpers to put it in there (or extend one if there's already such concept in Stencil, haven't checked)

Sources/Filters.swift
AliSoftware: Fair enough.

StencilSwiftKit.xcodeproj/xcshareddata/xcschemes/Tests.xcscheme
ggrell: Will fix, I must have checked it in inadvertently again

Sources/Filters.swift
djbe: @ggrell Any update on these last small items, or would you prefer I take them up?

AliSoftware:
I'm added the the small missing changes in #35, which will superseed this one so I'm closing it :wink:

AliSoftware:
This has now been merged by commit b2f414145dcbbe7eda5b011674f96fc6cdcd5bdd

AliSoftware:
Hey @ggrell !

Thanks a lot for contributing to SwiftGen & StencilSwiftKit! 👍
To thank you for your work and for taking time to improve the tool, I've invited you to join SwiftGen's GitHub organization's "CoreContributors" team :smiley:

No pressure to accept nor any obligation in that! It will just give you the ability to help triage issues or push directly branches to the repo (instead of your fork)… in case you want to contribute again in the future :wink:

ggrell:
Got overwhelmed with life lately so didn't have a chance to make those final changes; thank you for wrapping it up! I'd be happy to keep contributing though.

PR #25, opened 2017-02-24
Split rakefile into parts

djbe:
We should check how we can repeat these across repositories and keep them in sync. Right now there are some small differences due to layout differences and different requirements.


rakelib/utils.rake
```diff
@@ -0,0 +1,68 @@
+class Utils
+
+  # run a command using xcrun and xcpretty if applicable
+  def self.run(cmd, task, subtask = '', xcrun: false, xcpretty: false, direct: false)
```
AliSoftware: `xcpretty` and `direct` are mutually exclusive (i.e. the API technically allows you to call `xcpretty: true, direct: true` while it wouldn't make sense).

In a future version, maybe use a symbol instead, like `formatter: :xcpretty` vc `formatter: :none` vs `formatter: :direct`? (Although the difference between direct and none wouldn't be obvious so we'd need to find better names)

I'll let that small refactoring for the future PR in Eve consolidating rakelib accross repos though, so we can merge this now.

PR #28, opened 2017-03-01
Gemfile

djbe:
switch to using a gemfile and calling via bundle


rakelib/output.rake
```diff
@@ -0,0 +1,94 @@
...
+      Dir.glob("#{MODULE_OUTPUT_PATH}/*.swiftdoc").each do |f|
+        FileUtils.rm_rf(f)
+      end
+    end
+
+    desc 'Compile output'
+    task :compile => :modules do |task|
+      Utils.print_info 'Compiling template output files'
+
+      exit Dir.glob('Tests/Expected/**/*.swift').map { |f|
```
AliSoftware: `exit` typically takes an integer (the exit code) as argument, right?

So if one of the `compile_file(f, task)` returns `false`, the `reduce` will return `false` and `exit false` would be interpreted as `exit(0)`… generally meaning no error?
While if all `compile_file(f, task)` return `true`, the`reduce` will return `true` and `exit true` would be interpreted as `exit(1)`… meaning an error?

I think it would be better if `compile_file` returned the return code / status code of the execution of the `Utils.run` call (line 87), and doing a `exit_codes = Dir.glob(…).map { compile_file(…) }` first, then do a `return exit_codes.find { |c| c != 0 } || 0` to globally return the first non-zero exit code, or exit(0) if all exit codes are 0.

rakelib/output.rake
```diff
@@ -0,0 +1,94 @@
...
+        sdks = [:iphoneos, :macosx]
+      end
+
+      commands = sdks.map do |sdk|
+        %Q(--toolchain #{toolchain[:toolchain]} -sdk #{sdk} swiftc -parse -target #{SDKS[sdk]} -I #{toolchain[:module_path]} "#{MODULE_OUTPUT_PATH}/Definitions.swift" #{f})
+      end
+      subtask = File.basename(f, '.*')
+
+      begin
+        Utils.run(commands, task, subtask, xcrun: true)
```
AliSoftware: As stated above, `Utils.run` should hopefully return the status code of the ran command, so the same way as it's done on line 63, the function here should just end with the call to `Utils.run(…)` without the need of a `begin/rescue` block?

rakelib/pod.rake
```diff
@@ -1,7 +1,9 @@
-namespace :pod do
-  desc 'Lint the Pod'
-  task :lint do |task|
-    Utils.print_info 'Linting the pod spec'
-    Utils.run(%Q(pod lib lint "#{POD_NAME}.podspec" --quick), task)
+if File.file?('Podfile')
```
AliSoftware: Given the nature of the task, we should rather test the presence of a `#{PODNAME}.podspec` rather than the presence of a `Podfile`.

Rakefile
```diff
@@ -6,7 +6,6 @@ WORKSPACE = 'StencilSwiftKit'
 TARGET_NAME = 'Tests'
```
AliSoftware: Should we document somewhere (comment at the top of each `rakelib/*.rake` file? Elsewhere?) what global variables are used by various tasks, and thus which variables are expected to be defined in each repo's Rakefile?

djbe: Good idea.

rakelib/pod.rake
djbe: Will do

Rakefile
AliSoftware: Besides, `SCHEME_NAME` would be a better name for that variable 😉

rakelib/output.rake
djbe: Problem with that is that it can fail so we need to catch the exception.

rakelib/output.rake
djbe: The argument can be a boolean or an integer. If it's boolean, then true means success. If it's an integer than 0 means success.

AliSoftware: TIL 👍

PR #29, opened 2017-03-05
Ensure that the context enrich function doesn't overwrite existing variables

djbe:
Needed for https://github.com/SwiftGen/SwiftGenKit/pull/13

### DIFF
--- a/Sources/Context.swift
+++ b/Sources/Context.swift
@@ -9,8 +9,8 @@
 import Foundation
 
 public enum StencilContext {
-  public static let environment = "env"
-  public static let parameters = "param"
+  public static let environmentKey = "env"
+  public static let parametersKey = "param"
 
   /**
    Enriches a stencil context with parsed parameters and environment variables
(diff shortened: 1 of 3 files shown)


Sources/Context.swift
```diff
@@ -24,9 +24,19 @@ public enum StencilContext {
                             environment: [String: String] = ProcessInfo().environment) throws -> [String: Any] {
     var context = context
 
-    context[StencilContext.environment] = environment
-    context[StencilContext.parameters] = try Parameters.parse(items: parameters)
+    context[self.environment] = merge(context[self.environment], with: environment)
+    context[self.parameters] = merge(context[self.parameters], with: try Parameters.parse(items: parameters))
```
AliSoftware: Maybe we should rename those two `static let` constants using `environmentKey` and `parametersKey`.

Because the first time i read and reviewed your code change I didn't understand why you used instance properties there, and almost thought that `self.environment` were referring to the `[String: String]` dictionary of the context, containing the env variables (ok, doing the review from GitHub didn't help, as the limited git diff scope failed to remind me that this was actually  inside a `static` func…). Using `context[self.paramtersKey]` should help clarify that.

PR #32, opened 2017-04-25
Bump Stencil to version 0.9.0

diogot:
Also, update the project to Xcode 8.3.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ _None_
   [David Jennes](https://github.com/djbe)
   [#28](https://github.com/SwiftGen/SwiftGenKit/issues/28)
   [#31](https://github.com/SwiftGen/SwiftGenKit/issues/31)
-* Update Stecil to 0.9.0 and update project to Xcode 8.3.
+* Update Stencil to 0.9.0 and update project to Xcode 8.3.  
   [Diogo Tridapalli](https://github.com.diogot)
   [#32](https://github.com/SwiftGen/StencilSwiftKit/pull/32)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,10 +24,7 @@ _None_
   [David Jennes](https://github.com/djbe)
   [#28](https://github.com/SwiftGen/SwiftGenKit/issues/28)
   [#31](https://github.com/SwiftGen/SwiftGenKit/issues/31)
-* Update Stecil to 0.9.0
-  [Diogo Tridapalli](https://github.com.diogot)
-  [#32](https://github.com/SwiftGen/StencilSwiftKit/pull/32)
-* Update project to Xcode 8.3
+* Update Stecil to 0.9.0 and update project to Xcode 8.3.
   [Diogo Tridapalli](https://github.com.diogot)
   [#32](https://github.com/SwiftGen/StencilSwiftKit/pull/32)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,12 @@ _None_
   [David Jennes](https://github.com/djbe)
   [#28](https://github.com/SwiftGen/SwiftGenKit/issues/28)
   [#31](https://github.com/SwiftGen/SwiftGenKit/issues/31)
+* Update Stecil to 0.9.0
+  [Diogo Tridapalli](https://github.com.diogot)
+  [#32](https://github.com/SwiftGen/StencilSwiftKit/pull/32)
+* Update project to Xcode 8.3
+  [Diogo Tridapalli](https://github.com.diogot)
+  [#32](https://github.com/SwiftGen/StencilSwiftKit/pull/32)
 
 ## 1.0.1
 
--- a/circle.yml
+++ b/circle.yml
@@ -1,8 +1,6 @@
 machine:
   xcode:
     version: 8.3
-  pre:
-  - sudo launchctl load -w /System/Library/LaunchDaemons/com.apple.metadata.mds.plist
 
 dependencies:
   post:
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,6 @@
 machine:
   xcode:
-    version: 8.2
+    version: 8.3
 
 dependencies:
   post:
--- a/circle.yml
+++ b/circle.yml
@@ -1,6 +1,8 @@
 machine:
   xcode:
     version: 8.3
+  pre:
+  - sudo launchctl load -w /System/Library/LaunchDaemons/com.apple.metadata.mds.plist
 
 dependencies:
   post:
--- a/rakelib/utils.rake
+++ b/rakelib/utils.rake
@@ -68,6 +68,7 @@ class Utils
 
   # select the xcode version we want/support
   def self.version_select
+    Rake.sh 'mdimport /Applications/Xcode*'
     xcodes = `mdfind "kMDItemCFBundleIdentifier = 'com.apple.dt.Xcode' && kMDItemVersion = '8.*'"`.chomp.split("\n")
     if xcodes.empty?
       raise "\n[!!!] You need to have Xcode 8.x to compile SwiftGen.\n\n"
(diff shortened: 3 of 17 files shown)


circle.yml
```diff
@@ -1,6 +1,6 @@
 machine:
   xcode:
-    version: 8.2
+    version: 8.3
```
djbe: No idea if we should specify 8.3.2 because of this:
https://discuss.circleci.com/t/we-are-deprecating-the-xcode-8-3-build-image/12115

diogot: Looks like `8.3` already do that:
screen shot 2017-04-25 at 08 07 10
From https://circleci.com/gh/SwiftGen/StencilSwiftKit/119

CHANGELOG.md
```diff
@@ -24,6 +24,12 @@ _None_
   [David Jennes](https://github.com/djbe)
   [#28](https://github.com/SwiftGen/SwiftGenKit/issues/28)
   [#31](https://github.com/SwiftGen/SwiftGenKit/issues/31)
+* Update Stecil to 0.9.0
```
djbe: Please add a `.` and 2 spaces at the end of the description. Also, join these 2 changes together.

circle.yml
djbe: Ah, cool, didn't check the CI output.

It's weird though that our script isn't finding Xcode.

diogot: Although [this](https://github.com/SwiftGen/StencilSwiftKit/blob/master/rakelib/utils.rake#L71) is preventing the tests to run on [CircleCI](https://circleci.com/gh/SwiftGen/StencilSwiftKit/120).
Have any suggestion?

djbe: I had the same issue locally and had to reset Spotlight for the applications folder. No idea how to fix this on CI.

@AliSoftware or @DanToml: any thoughts?

diogot: Maybe I should divide in two PRs

djbe: Nah, wait until we can find a cause (and hopefully solution).

AliSoftware: See also https://mobile.twitter.com/abizern/status/852120220328218625 (= we are not alone and it's a general issue, as latest Xcode seems to be excluded from spotlight by default) and https://github.com/KrauseFx/xcode-install/issues/204 (discussion about possible alternatives)

djbe: Spotlight is just completely broken on CI, nothing I try with `mdfind` works.

Maybe we should change our rakelib so that, when we can't find Xcode, we default to `/Applications/Xcode.app`. If we can't find that, then we throw an error.

The possible alternatives aren't really viable: system_profiler depends on spotlight, and pkgutil is too slow.

diogot: Did you know [xcenv](https://github.com/xcenv/xcenv)?

diogot: They also use [Spotlight](https://github.com/xcenv/xcenv#finding-xcodeapp)

AliSoftware: Maybe one other fallback before directly selecting `/Applications/Xcode.app` (which isn't necessary the one we want) is to `ls -1 /Applications | grep '^Xcode*.app'`, then use `plutil` to check the `CFBundleIdentifier` and extract the `CFBundleVersion` ourselves. We be a little like re-creating `mdls` manually but shouldn't be too much work and probably more reliable.

CHANGELOG.md
djbe: You're still missing the 2 spaces after the dot.

CHANGELOG.md
```diff
@@ -24,6 +24,9 @@ _None_
   [David Jennes](https://github.com/djbe)
   [#28](https://github.com/SwiftGen/SwiftGenKit/issues/28)
   [#31](https://github.com/SwiftGen/SwiftGenKit/issues/31)
+* Update Stecil to 0.9.0 and update project to Xcode 8.3.
```
AliSoftware: s/Stecil/Stencil/

circle.yml
diogot: Maybe [`mdimport /Applications/Xcode*`](http://osxdaily.com/2012/02/02/reindex-spotlight-from-the-command-line/) and retry.

diogot: Looks like [this](https://discuss.circleci.com/t/announcing-xcode-8-3-support-and-opt-in-upgrades-for-macos/11528/21) fixed.

AliSoftware: Cool, that's a good first step 😉 Unfortunately it will only solve it on CircleCI (which is a good start) but not locally.

The `mdimport /Applications/Xcode*` might be worth a try for local builds

diogot: I need a combination of both, but it's working now.

CHANGELOG.md
diogot: Sorry

rakelib/utils.rake
```diff
@@ -68,6 +68,7 @@ class Utils
 
   # select the xcode version we want/support
   def self.version_select
+    Rake.sh 'mdimport /Applications/Xcode* && sleep 1'
```
AliSoftware: I really don't like that we have to hardcode an arbitrary delay there. Slows down the build and duration might vary anyway depending on machine and how many Xcode bundles the command are gonna indeed… I'm sure we can do better?

diogot: I agree that is far from ideal, but [this](https://circleci.com/gh/SwiftGen/StencilSwiftKit/128) run failed `rake xcode:test` yet passed `rake spm:test`, indicating that it might have a delay between indexing and get it available to `mdfind`.

Another approach might be run it before, in `lint:install` for example.

Any ideas?

circle.yml
DanToml: Ah we have spotlight indexing disabled by default, good call on the `mdimport` @diogot.

rakelib/utils.rake
djbe: What @AliSoftware means is that an arbitrary sleep might not fix it on all systems, as we never know how long indexing will take. 

I think it might be better to write our own fallback search in case spotlight doesn't do it for us. It doesn't have to be extensive, we can always fail and tell the user the reason, and it's up to them to then fix spotlight on their device.

Another thing is that with this line, the `version_select` call will always at least take 1s to execute, slowing down the whole build process. Might be best to make this somehow static (like a swift static constant, my ruby-fu is failing me right now 😆 ) to cache the result.

diogot: 🤔, I see.

I agree that `sleep 1` is a hack that won't fix the problem in a definitive way, and it looks like this spotlight problem is a bug on [10.12.4](https://openradar.appspot.com/31646293).

Enable spotlight was the correct fix for CircleCI (as @DanToml mentioned), so we still need a workaround to Sierra bug. It's plausible to expect that the next Sierra update will fix this bug, so I think that might be unnecessary to have a fallback method to spotlight that will be necessary just for a couple of months.

So I have a proposal that would be more assertive, what if remove the sleep and update the error message to something like: 

> SwiftGen requires Xcode 8.x, but we were not able not found it. If it's already installed update your Spotlight index with "mdimport /Applications/Xcode*"

It will work reliable for most of the cases without any hacks and we inform the user a way to fix if an error happens.

What do you think?

djbe: That's actually ok for me. It avoids weird hacks and development of something that might not be needed in a few weeks/months (hopefully).

AliSoftware: :+1:

diogot:
Is there anything else that need attention?

djbe:
Nope, LGTM to me.

diogot:
Nice! Any plans for a release?
I did the bump for SwiftGen too, but I think it's better to point to a release instead of a commit.

AliSoftware:
I've been super-busy with conferences and work and personal stuff lately, but after my talk/live-coding this afternoon (in just 1 hour in NSBudapest :wink: ) I'll probably have a little more time (at last I hope) to start catching up with all the pending PRs and try to make that happen as soon as possible.

No promises of any exact timeline, as things have been pretty crazy lately and I just hope I'll have enough bandwidth, but I don't want to wait any longer anyway so I'll do everything I can to make it happen soon!

diogot:
Nice, good luck with the talk/live-coding 😀

PR #36, opened 2017-04-30
Split "filters" files

AliSoftware:
1. Make files better organized, changing struct namespaces to enums and renaming some files
2. Also move types into nested types to make it more swifty and better namespaced.
Old names for those types still exists as typealiases but marked as deprecated and will be removed by #37 
3. Also uniformized documentation in the process


Sources/Parameters.swift
```diff
@@ -6,16 +6,34 @@
...
+  public enum Error: Swift.Error {
+    case invalidSyntax(value: Any)
+    case invalidKey(key: String, value: Any)
+    case invalidStructure(key: String, oldValue: Any, newValue: Any)
+  }
+
   typealias Parameter = (key: String, value: Any)
   public typealias StringDict = [String: Any]
 
+  /// Transforms a list of strings representing structred-key/value pairs, like
```
djbe: structured

Sources/Parameters.swift
```diff
@@ -6,16 +6,34 @@
...
+  /// `["pt.x=1", "pt.y=2", "values=1", "values=2", "values=3" "flag"]`
```
djbe: missing comma before flag

Sources/Parameters.swift
```diff
@@ -6,16 +6,34 @@
...
+  /// into a structured dictionary.
+  ///
+  /// - Parameter items: The list of `k=v`-style Strings, each string
```
djbe: Shouldn't we use the same structure as in the other doc. blocks? Namely a `- Paramters:`, then newline, then a list of parameters (here `- items:`).

Sources/Parameters.swift
```diff
@@ -33,13 +51,23 @@ public enum Parameters {
     }
   }
 
+  // MARK: - Private methods
+
+  /// Parse a single `key=value` (or `key`) string and inserts it into
+  /// an existing StringDict dictionary being built.
+  ///
+  /// - Parameters:
+  ///   - parameter: The parameter/string (key/Value pair) to parse
```
djbe: Key or value (case)

Sources/Parameters.swift
```diff
@@ -33,13 +51,23 @@ public enum Parameters {
...
+  ///   - result: The dictionary currently being built and to which to add the value
+  /// - Returns: The new content of the dictionary being built after nserting the new parsed value
```
djbe: inserting

Sources/Filters+Strings.swift
```diff
@@ -0,0 +1,169 @@
...
+      result = camelCased.stringByReplacingMatches(in: result,
+                                                   options: .reportCompletion,
+                                                   range: fullRange,
+                                                   withTemplate: "$1_$2")
+      return result.replacingOccurrences(of: "-", with: "_")
+    }
+
+    /// Checks if the string is one of the reserved keywords and if so, escapes it using backticks
+    ///
+    /// - Parameter in: the string to possibly escape
```
djbe: Use same structure as other doc blocks:
```
 - Parameters:
   - in: ...
```

Sources/Filters+Strings.swift
```diff
@@ -0,0 +1,169 @@
...
+
+    private static func containsAnyLowercasedChar(_ string: String) throws -> Bool {
+      let lowercaseCharRegex = try NSRegularExpression(pattern: "[a-z]", options: .dotMatchesLineSeparators)
+      let fullRange = NSRange(location: 0, length: string.unicodeScalars.count)
+      return lowercaseCharRegex.firstMatch(in: string, options: .reportCompletion, range: fullRange) != nil
+    }
+
+    /// This returns the snake cased variant of the string.
+    ///
+    /// - Parameter string: The string to snake_case
```
djbe: See comments below related to doc block structure.

Sources/Filters+Strings.swift
```diff
@@ -0,0 +1,169 @@
...
+    static func escapeReservedKeywords(value: Any?) throws -> Any? {
+      guard let string = value as? String else { throw Filters.Error.invalidInputType }
+      return escapeReservedKeywords(in: string)
+    }
+
+    // MARK: - Private methods
+
+    /// This returns the string with its first parameter uppercased.
+    /// - note: This is quite similar to `capitalise` except that this filter doesn't
+    ///          lowercase the rest of the string but keep it untouched.
```
djbe: keeps

Sources/Parameters.swift
AliSoftware: Well when you ask Xcode to generate the doc comment (menu "Editor" > "Structure" > "Add Documentation"), it uses the "`- Parameters` + one param per line" if there's more than one parameter, but uses directly `- Parameter foo` if there's only one parameter.

I agree that's inconsistent, and I actually wondered the same as you here. idk if it's better to force consistency ourselves, using `- Parameters` + one param per line, or to just do as Xcode does (which at least would not force us to fight Xcode and correct every new contributor who are using Xcode's doc auto-generation shortcut)

djbe: Hmm, leave it then. (I didn't even know that functionality existed).

AliSoftware: @djbe  Approved for merge, then?

djbe: Jup!

Issue #5, opened 2017-01-04
Cleanup (remove) old stuff


djbe:

djbe:
Fixed by #10 and #37.

PR #37, opened 2017-04-30
Remove deprecated, non nested error types

AliSoftware:
This PR sits on top of #36 and is breaking.
It can be considered one small part of #5

It should only be merged once #36 has been merged then released as a new version on CP. Then we could rebase it on top of the new master after release and merge it once we're ready for the next major version to removed the depreciations.

Note that we'll probably have a conflict on the CHANGELOG when rebasing and during that conflict we should make sure to move the entries to the (future) proper section associated with the right version and not to keep them at their current location.

TODO:
 - [x] Rebase on master
 - [x] Check that changelog entry is in correct position after merge.


AliSoftware:
#36 has now been merged and this branch rebased on top of master after the release of 1.0.2
So it's ready for the next major now

PR #41, opened 2017-05-16
Filters with parameters

### LINKED ISSUE #6: Migrate some filters to use parameters
Some Filters were created at a time when Stencil didn't support filter parameters yet.

As a result, we have some filters that have been duplicated to allow some variations, but now that Stencil filters support parameters, these duplications can be removed and changed to a filter parameter instead.

One typical example is filters `StringFilters.snakeToCamelCase` + `StringFilters.snakeToCamelCaseNoPrefix` that can be unified to a single `StringFilters.snakeToCamelCase` filter with a parameter (`prefix: Bool`)

djbe:
Fixes #6.

Unless I'm missing something, the only filter we can change this way is the `snakeToCamelCase` filter.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@
 * [String filters](Documentation/filters-strings.md):
   * `escapeReservedKeywords`: Escape keywods reserved in the Swift language, by wrapping them inside backticks so that the can be used as regular escape keywords in Swift code.
   * `lowerFirstWord`
-  * `snakeToCamelCase` / `snakeToCamelCaseNoPrefix`
+  * `snakeToCamelCase`: Transforms text from snake_case to camelCase. By default it keeps prefix underscores, unless a single optional argument is set to "true", "no" or "0".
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -27,7 +27,7 @@
 * [String filters](Documentation/filters-strings.md):
   * `escapeReservedKeywords`: Escape keywods reserved in the Swift language, by wrapping them inside backticks so that the can be used as regular escape keywords in Swift code.
   * `lowerFirstWord`
-  * `snakeToCamelCase` / `snakeToCamelCaseNoPrefix`
+  * `snakeToCamelCase`: Transforms text from snake_case to camelCase. By default it keeps prefix underscores, unless a single optional argument is set to "true", "no" or "0".
```
AliSoftware: s/prefix/leading/ ?

djbe: sure

AliSoftware: Once fixed, good to merge 👍

PR #42, opened 2017-05-17
Provide migration guide

djbe:
- [x] Merge #41 first
- [x] Add link from changelog to migration guide

### DIFF
--- a/Documentation/MigrationGuide.md
+++ b/Documentation/MigrationGuide.md
@@ -1,0 +1,21 @@
+## StencilSwiftKit 2.0 (SwiftGen 5.0) ##
+
+### For template writers:
+
+* We've removed our `join` array filter as Stencil provides it's own version that accepts a parameter. If you were using StencilSwiftKit's version, replace instances of:  
+```{{ myArray|join }}```  
+with:  
+```{{ myArray|join:", " }}```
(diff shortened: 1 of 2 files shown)


Documentation/MigrationGuide.md
```diff
@@ -0,0 +1,21 @@
...
+* We've removed our `join` array filter as Stencil provides it's own version that accepts a parameter. If you were using StencilSwiftKit's version, replace instances of:  
+```{{ myArray|join }}```  
+with:  
+```{{ myArray|join:", " }}```
+* We've refactored our `snakeToCamelCase` to accept arguments, thus removing the need for `snakeToCamelCaseNoPrefix`. If you were using the latter, replace instances of:  
+```{{ myValue|snakeToCamelCaseNoPrefix }}```  
+with:  
+```{{ myValue|snakeToCamelCase:true }}```
+
+### For developers:
```
AliSoftware: "### For developers using StencilSwiftKit as a Swift dependency" maybe? (Or if that makes the title too long, at least explain that in the first line/paragraph of that section)

Because most people using SwiftGen are writing code as a daily job and consider themselves as developers and might think that section applies to them 😉

djbe: Hehe, I know 😆 I'll add that (without the Swift part, that should be clear enough), clarity is more important than nice titles.

PR #45, opened 2017-05-19
Fix snakeToCamelCase readme documentation

Liquidsoul:
I've noticed that the information about `snakeToCamelCase`'s parameter in the README is wrong so this PR fixes this.
I've also added some tests to show that and ensure that I was not mistaking 😅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@
 
 ### Bug Fixes
 
-_None_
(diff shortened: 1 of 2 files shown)


djbe:
Should we repeat those tests though? Aren't those cases already covered by the `parseBool` tests?

Liquidsoul:
Maybe not. I'll remove the added tests in the PR.

PR #48, opened 2017-05-30
Filter: Only remove multiple spaces

djbe:
Modify the `removeNewlines` filter to only remove whitespace characters:
- if they're a newline
- if there are multiple consecutive whitespace chars
- if they're at either end of the string (trimming)

Remaining single whitespace characters are converted to a space ` `.

### DIFF
--- a/Documentation/filters-strings.md
+++ b/Documentation/filters-strings.md
@@ -49,23 +49,27 @@ Transforms an arbitrary string so that only the first "word" is lowercased.
 
 ## Filter: `removeNewlines`
 
-Removes all newlines and whitespace characters from the string.
+This filter has a couple of modes that you can specifiy using an optional argument (defaults to "all"):
 
-| Input                 | Output                |
-|-----------------------|-----------------------|
-| `  \ntest`            | `test`                |
-| `test \n\t `          | `test`                |
-| `test\n  test`        | `testtest`            |
-| `\r\ntest\n  test\n`  | `testtest`            |
+**all**: Removes all newlines and whitespace characters from the string.
 
-By default it removes whitespace characters, unless a single optional argument is set to "false", "no" or "0":
+| Input                  | Output                |
+|------------------------|-----------------------|
+| `  \ntest`             | `test`                |
+| `test \n\t `           | `test`                |
+| `test\n  test`         | `testtest`            |
+| `test, \ntest, \ntest` | `test,test,test`      |
+| ` test test `          | `testtest`            |
 
-| Input                 | Output                |
-|-----------------------|-----------------------|
-| `  \ntest`            | `  test`              |
-| `test \n\t `          | `test \t `            |
-| `test\n  test`        | `test  test`          |
-| `\r\ntest\n  test\n`  | `test  test`          |
+**leading**: Removes leading whitespace characters from each line, and all newlines. Also trims the end result.
+
+| Input                  | Output                |
+|------------------------|-----------------------|
+| `  \ntest`             | `test`                |
+| `test \n\t `           | `test`                |
+| `test\n  test`         | `testtest`            |
+| `test, \ntest, \ntest` | `test, test, test`    |
+| ` test test `          | `test test`           |
 
 ## Filter: `snakeToCamelCase`
 
--- a/Sources/Filters+Strings.swift
+++ b/Sources/Filters+Strings.swift
@@ -71,7 +71,7 @@ extension Filters {
     /// - Returns: the camel case string
     /// - Throws: FilterError.invalidInputType if the value parameter isn't a string
     static func snakeToCamelCase(_ value: Any?, arguments: [Any?]) throws -> Any? {
-      let stripLeading = try Filters.parseBool(from: arguments, index: 0, required: false) ?? false
+      let stripLeading = try Filters.parseBool(from: arguments, required: false) ?? false
       guard let string = value as? String else { throw Filters.Error.invalidInputType }
 
       let unprefixed: String
@@ -104,7 +104,7 @@ extension Filters {
     /// - Returns: the snake case string
     /// - Throws: FilterError.invalidInputType if the value parameter isn't a string
     static func camelToSnakeCase(_ value: Any?, arguments: [Any?]) throws -> Any? {
-      let toLower = try Filters.parseBool(from: arguments, index: 0, required: false) ?? true
+      let toLower = try Filters.parseBool(from: arguments, required: false) ?? true
       guard let string = value as? String else { throw Filters.Error.invalidInputType }
 
       let snakeCase = try snakecase(string)
--- a/Sources/Filters+Strings.swift
+++ b/Sources/Filters+Strings.swift
@@ -7,6 +7,10 @@
 import Foundation
 import Stencil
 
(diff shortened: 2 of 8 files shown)


Documentation/filters-strings.md
```diff
@@ -57,15 +57,17 @@ Removes all newlines and whitespace characters from the string.
...
-By default it removes whitespace characters, unless a single optional argument is set to "false", "no" or "0":
+By default it removes consecutive whitespace characters (and trims the ends), unless a single optional argument is set to "false", "no" or "0":
 
 | Input                 | Output                |
 |-----------------------|-----------------------|
 | `  \ntest`            | `  test`              |
 | `test \n\t `          | `test \t `            |
 | `test\n  test`        | `test  test`          |
 | `\r\ntest\n  test\n`  | `test  test`          |
+| ` test test `         | ` test test `         |
```
AliSoftware: Not sure why the beginning & end whitespaces would be kept there? I'd expect them to be removed by `.trimmingCharacters(in: .whitespacesAndNewlines)` here.

Tests/StencilSwiftKitTests/StringFiltersTests.swift
```diff
@@ -140,11 +140,12 @@ extension StringFiltersTests {
...
-      "  \n test": "test",
-      "test  \n ": "test",
-      "test\n \ntest": "testtest",
-      "\n test\n \ntest \n ": "testtest"
+      "test1": "test1",
+      "  \n test2": "test2",
+      "test3  \n ": "test3",
+      "test4\n \ntest": "test4test",
+      "\n test5\n \ntest test \n ": "test5test test",
+      "test6\ntest": "test6test"
```
AliSoftware: Maybe add that `" test test ":"test test"` here to confirm my suggestion about that it should remove the leading/trailing whitespaces in the examples you put in the md docs above

Documentation/filters-strings.md
djbe: This is for the false case, where only newlines are removed.This is

AliSoftware: Oh, right.

AliSoftware: Oh, right.

djbe:
Hmmm, instead of doing multiple regexes, would it be easier (for the removewhitespaces case) to:
- split based on newlines
- trim each line
- join again

Tests/StencilSwiftKitTests/StringFiltersTests.swift
AliSoftware: Still, I like that the examples we give in the MD documentation are used in the tests as well, if only to be sure that we are consistent with what we state in the doc and that those cases are indeed checked

djbe: I'll add that specific case, but I'll number it (easier for debugging if needed) like the rest.

AliSoftware:
As long as the test cases still passes when you change the implementation… 😉 
I'm not sure what would be faster, but RegExes even if powerful are also generally slower indeed, so might be worth it yeah

djbe:
I'm just wondering what would be useful for users (and our use cases).

The main one will be the ability to write things like macros on multiple lines but still have the result be one line. This might mean multiple things though:
- I don't want any spaces at all, anywhere
- Just remove the newlines and leading whitespaces
- ...?

AliSoftware:
That's pretty much the use cases I see indeed. Creating some identifier or word from a complex expression (= no spaces at all anywhere) or formatting the template nicely (remove newlines and leading spaces per lines), so I think this covers all we need for now?

djbe:
Hmmm no then I need to change the code a bit. Right now it either:
- only removes newlines and nothing else
- removes newlines and multiple consecutive spaces + replaces single whitespaces with a space

djbe:
Instead of a boolean parameter, should we maybe accept a string to represent a bunch of cases ("all", "leading", ...)

AliSoftware:
Mmmh I was gonna say YAGNI, but given that the different cases are not that straightforward to explain ("remove whitespaces except this but that…"), a string instead of a bool might be more explicit anyway, so yeah, go for it.

Sources/Filters+Strings.swift
```diff
@@ -120,13 +120,23 @@ extension Filters {
     }
 
     static func removeNewlines(_ value: Any?, arguments: [Any?]) throws -> Any? {
-      let removeSpaces = try Filters.parseBool(from: arguments, index: 0, required: false) ?? true
+      let mode = arguments.first as? String ?? "all"
```
AliSoftware: Shouldn't we throw if the argument is not a valid string, like `removeNewlines:42`?

So if there's no argument at all (`arguments.first == nil`), agreed that `all` should be the default. But if the argument isn't castable as a `as? String`, maybe throw?

Sources/Filters+Strings.swift
```diff
@@ -120,13 +120,23 @@ extension Filters {
...
       guard let string = value as? String else { throw Filters.Error.invalidInputType }
 
-      let set: CharacterSet = removeSpaces ? .whitespacesAndNewlines : .newlines
-      let result = string.components(separatedBy: set).joined()
-
-      return result
+      switch mode {
+      case "all":
```
AliSoftware: Use some constants instead of magic strings here? Even if those constants are scoped inside the func.

e.g.
```swift
static func removeNewlines(_ value: Any? arguments: [Any?]) throws -> Any? {
  enum Modes: String {
    case all, leading
  }

  let modeArg = String(describing: arguments.first) ?? Modes.all.rawValue
  guard let mode = Modes(rawValue: modeArg) else {
    throw Filters.Error.invalidOption(option: modeArg)
  }

  switch mode {
  case .all: …
  case .leading: …
  }
}
```

djbe:
Before you ask, the reason that enum is not inside the function is because `swiftlint` complains that such things can be at most 1 level deep (nesting) 🤷‍♂️

Sources/Filters+Strings.swift
```diff
@@ -119,14 +123,31 @@ extension Filters {
...
-      return result
+      switch mode {
+      case .all:
+        return string
+          .components(separatedBy: .whitespacesAndNewlines)
+          .joined()
+      case .leading:
+        return string
+          .components(separatedBy: .newlines)
+          .map { String($0.unicodeScalars.drop(while: { CharacterSet.whitespaces.contains($0) })) }
```
AliSoftware: Codebeat seems to be unhappy with nesting too deep here, maybe separate the closure used in that map in a private function to make it happy?

Tests/StencilSwiftKitTests/ParseEnumTests.swift
```diff
@@ -0,0 +1,41 @@
...
+    let value = try Filters.parseEnum(from: ["foo"], default: Test.baz)
+    XCTAssertEqual(value, Test.foo)
+  }
+
+  func testParseEnum_WithBarString() throws {
+    let value = try Filters.parseEnum(from: ["bar"], default: Test.baz)
+    XCTAssertEqual(value, Test.bar)
+  }
+
+  func testParseEnum_WithBazString() throws {
```
AliSoftware: Add a test with a non-zero index argument too

PR #52, opened 2017-05-31
Refactor the swiftidentifier code to use a ranges array

djbe:
Let's see if codebeat sees this.

### DIFF
--- a/Sources/SwiftIdentifier.swift
+++ b/Sources/SwiftIdentifier.swift
@@ -6,82 +6,51 @@
 
 import Foundation
 
+private func mr(_ char: Int) -> CountableClosedRange<Int> {
+  return char...char
+}
+
 // Official list of valid identifier characters
 // swiftlint:disable:next line_length
 // from: https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/LexicalStructure.html#//apple_ref/doc/uid/TP40014097-CH30-ID410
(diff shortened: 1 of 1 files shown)


Sources/SwiftIdentifier.swift
```diff
@@ -6,82 +6,51 @@
...
+private func mr(_ char: Int) -> CountableClosedRange<Int> {
+  return char...char
+}
+
 // Official list of valid identifier characters
 // swiftlint:disable:next line_length
 // from: https://developer.apple.com/library/ios/documentation/Swift/Conceptual/Swift_Programming_Language/LexicalStructure.html#//apple_ref/doc/uid/TP40014097-CH30-ID410
-// swiftlint:disable function_body_length
+private let headRanges: [CountableClosedRange<Int>] = [
+  0x61...0x7a, 0x41...0x5a, mr(0x5f), mr(0xa8), mr(0xaa), mr(0xad), mr(0xaf),
```
AliSoftware: Nitpicking Olivier strikes again: why is `0x61...0x7a` now _before_ `0x41...0x5a`?

djbe: Because I copied from https://github.com/apple/swift-protobuf/blob/master/Sources/PluginLibrary/SwiftLanguage.swift

AliSoftware: Ah! Right then!

PR #56, opened 2017-07-29
Fix compilation on Linux

jpsim:


jpsim:
I briefly started looking into making Sourcery build and run on Linux and this is the only dependency that doesn't support it (other than SwiftTryCatch, which isn't a strong requirement for Sourcery). Minimal changes required.

Sources/Context.swift
```diff
@@ -22,7 +22,8 @@ public enum StencilContext {
   /// - Throws: `Parameters.Error`
   public static func enrich(context: [String: Any],
                             parameters: [String],
-                            environment: [String: String] = ProcessInfo().environment) throws -> [String: Any] {
+                            environment: [String: String] =
+                              ProcessInfo.processInfo.environment) throws -> [String: Any] {
```
Liquidsoul: You need to either remove two spaces here or move this code back to the previous line and disable the `line_length` rule just this one to allow the CI to pass.

jpsim: That's a SwiftLint bug. I've filed a PR to fix it here: realm/SwiftLint#1726

However, since that patch won't be available until a future SwiftLint release, I'll adapt my code to pass linting with the current release of SwiftLint.

Liquidsoul:
LGTM, thanks a lot for the changes!
We will try to make a release ASAP so you can continue your work on Sourcery on Linux 😉

AliSoftware:
@jpsim Released 2.0.1 with your fix 👍 

```
--------------------------------------------------------------------------------
 🎉  Congrats

 🚀  StencilSwiftKit (2.0.1) successfully published
 📅  August 12th, 18:17
 🌎  https://cocoapods.org/pods/StencilSwiftKit
 👍  Tell your friends!
--------------------------------------------------------------------------------
```

PR #61, opened 2017-08-17
Pretty swiftidentifier

### LINKED ISSUE #30: Identifiers starting with numbers when combining `swiftIdentifier` and `snakeToCamelCase`
Generated output example:
...enum UniversLTStd: String, FontConvertible {
    case 65Bold = "UniversLTStd-Bold"
    case 67BoldCondensed = "UniversLTStd-BoldCn"
    case 57Condensed = "UniversLTStd-Cn"
    case 47LightCondensed = "UniversLTStd-LightCn"
    case 47LightCondensedOblique = "UniversLTStd-LightCnObl"
    case 59UltraCondensed = "UniversLTStd-UltraCn"
  }...

djbe:
Fixes #30 (refs https://github.com/SwiftGen/SwiftGen/issues/289).

### DIFF
--- a/Documentation/filters-strings.md
+++ b/Documentation/filters-strings.md
@@ -184,16 +184,21 @@ https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift
 | `some$URL`             | `Some_URL`              |
 | `25 Ultra Light`       | `_25_Ultra_Light`       |
 | `26_extra_ultra_light` | `_26_extra_ultra_light` |
+| `apples.count`         | `Apples_Count`          |
+| `foo_bar.baz.qux-yay`  | `Foo_bar_Baz_Qux_Yay`   |
 
-**pretty**: Same as normal, but it will first replace whitespaces by underscores and apply the `snakeToCamelCase` filter.
+**pretty**: Same as normal, but afterwards it will apply the `snakeToCamelCase` filter, and other manipulations, for a prettier (but still valid) identifier.
 
 | Input                  | Output               |
 |------------------------|----------------------|
 | `hello`                | `Hello`              |
 | `42hello`              | `_42hello`           |
-| `some$URL`             | `Some_URL`           |
+| `some$URL`             | `SomeURL`            |
 | `25 Ultra Light`       | `_25UltraLight`      |
 | `26_extra_ultra_light` | `_26ExtraUltraLight` |
+| `apples.count`         | `ApplesCount`        |
+| `foo_bar.baz.qux-yay`  | `FooBarBazQuxYay`    |
+
 
 ## Filter: `titlecase`
 
--- a/Documentation/filters-strings.md
+++ b/Documentation/filters-strings.md
@@ -166,7 +166,9 @@ This filter accepts a parameter (boolean, default `false`) that controls the pre
 
 ## Filter: `swiftIdentifier`
 
-Transforms an arbitrary string into a valid Swift identifier (using only valid characters for a Swift identifier as defined in the Swift language reference). It will apply the following rules:
+This filter has a couple of modes that you can specifiy using an optional argument (defaults to "normal"):
+
+**normal**: Transforms an arbitrary string into a valid Swift identifier (using only valid characters for a Swift identifier as defined in the Swift language reference). It will apply the following rules:
 
 - Uppercase the first character.
 - Prefix with an underscore if the first character is a number.
@@ -175,11 +177,23 @@ Transforms an arbitrary string into a valid Swift identifier (using only valid c
 The list of allowed characters can be found here:
 https://developer.apple.com/library/content/documentation/Swift/Conceptual/Swift_Programming_Language/LexicalStructure.html
 
-| Input      | Output     |
-|------------|------------|
-| `hello`    | `Hello`    |
-| `42hello`  | `_42hello` |
-| `some$URL` | `Some_URL` |
+| Input                  | Output                  |
+|------------------------|-------------------------|
+| `hello`                | `Hello`                 |
+| `42hello`              | `_42hello`              |
+| `some$URL`             | `Some_URL`              |
+| `25 Ultra Light`       | `_25_Ultra_Light`       |
+| `26_extra_ultra_light` | `_26_extra_ultra_light` |
+
+**pretty**: Same as normal, but it will first replace whitespaces by underscores and apply the `snakeToCamelCase` filter.
+
+| Input                  | Output               |
+|------------------------|----------------------|
+| `hello`                | `Hello`              |
+| `42hello`              | `_42hello`           |
+| `some$URL`             | `Some_URL`           |
+| `25 Ultra Light`       | `_25UltraLight`      |
+| `26_extra_ultra_light` | `_26ExtraUltraLight` |
 
 ## Filter: `titlecase`
 
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@
   * `removeNewlines`: Removes newlines and other whitespace characters, depending on the mode ("all" or "leading").
   * `replace`: Replaces instances of a substring with a new string.
   * `snakeToCamelCase`: Transforms text from snake_case to camelCase. By default it keeps leading underscores, unless a single optional argument is set to "true", "yes" or "1".
-  * `swiftIdentifier`: Transforms an arbitrary string into a valid Swift identifier (using only valid characters for a Swift identifier as defined in the Swift language reference)
+  * `swiftIdentifier`: Transforms an arbitrary string into a valid Swift identifier (using only valid characters for a Swift identifier as defined in the Swift language reference). In "pretty" mode, it will also first apply the snakeToCamelCase filter.
   * `upperFirstLetter`: Uppercases only the first character
 * [Number filters](Documentation/filters-numbers.md):
   * `int255toFloat`
--- a/README.md
+++ b/README.md
@@ -36,7 +36,7 @@
   * `removeNewlines`: Removes newlines and other whitespace characters, depending on the mode ("all" or "leading").
   * `replace`: Replaces instances of a substring with a new string.
   * `snakeToCamelCase`: Transforms text from snake_case to camelCase. By default it keeps leading underscores, unless a single optional argument is set to "true", "yes" or "1".
-  * `swiftIdentifier`: Transforms an arbitrary string into a valid Swift identifier (using only valid characters for a Swift identifier as defined in the Swift language reference). In "pretty" mode, it will also first apply the snakeToCamelCase filter.
+  * `swiftIdentifier`: Transforms an arbitrary string into a valid Swift identifier (using only valid characters for a Swift identifier as defined in the Swift language reference). In "pretty" mode, it will also apply the snakeToCamelCase filter afterwards, and other manipulations if needed for a "prettier" but still valid identifier.
   * `upperFirstLetter`: Uppercases only the first character
 * [Number filters](Documentation/filters-numbers.md):
   * `int255toFloat`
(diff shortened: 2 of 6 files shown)


djbe:
This still doesn't achieve what we want. I've just finished modifying our templates to use `swiftIdentifier:"pretty"` instead of `swiftIdentifier|snakeToCamelCase`, but we didn't account for these cases:
- `Cyan-Color` --> `Cyan_Color` instead of `CyanColor`
- `.SFNSDisplay` --> `_SFNSDisplay` instead of `SFNSDisplay`
- `.SFNSDisplay` --> `_SFNSDisplay` instead of `SFNSDisplay`
- `Accept-CGU` --> `Accept_CGU` instead of `AcceptCGU`
- `Show-NavCtrl` --> `Show_NavCtrl` instead of `ShowNavCtrl`
- `apples.count` --> `Apples_Count` instead of `ApplesCount`
- `multiLine\nKey` --> `MultiLine_Key` instead of `MultiLineKey`

Replacing the current "pretty" implementation from:
`replace:" "|snakeToCamelCase:"true"|swiftIdentifier`
with:
`swiftIdentifier|snakeToCamelCase:"true"|swiftIdentifier`
solves all the issues mentioned above.

@AliSoftware I know you didn't want to run the filter twice, how would you suggest solving these issues?

AliSoftware:
Replace not only spaces but also dashes and periods with underscore before snakeToCamelCase? Or more generally using a RegEx and use the word separator character class to split the string first?

djbe:
Right, replace `-` and `.`. But what about `\n`? (see multiline key) Should the user expect those to be replaced too? Or other characters?

AliSoftware:
We could also use `string.enumerateSubstringsInRange(range, options: NSStringEnumerationOptions.ByWords)`

AliSoftware:
> Right, replace - and .. But what about \n? (see multiline key) Should the user expect those to be replaced too? Or other characters?

That's why I mentioned the more generic answer of using some official unicode word separator 😉

djbe:
I have to admit I don't know what that matches (words, sure, but how?). Going to read some docs 😄

AliSoftware:
Or `CFStringTokenizer` and `kCFStringTokenizerUnitWord`.
Or `NSLinguisticTagger`

(whatever is compatible with Linux would be preferable)

AliSoftware:
Quick Playground test:

```swift

import Foundation

extension String {
  func words() -> [String] {

    let range = self.startIndex..<self.endIndex
    var words = [String]()

    self.enumerateSubstrings(in: range, options: .byWords) { (substring, _, _, _) -> () in
      words.append(substring!)
    }

    return words
  }

  func upFirst() -> String {
    guard let first = self.characters.first else { return self }
    return String(first).uppercased() + String(self.characters.dropFirst())
  }
}

extension Array where Element == String {
  func camelJoin() -> String {
    return self.map { $0.upFirst() }.joined(separator: "")
  }
}


print("Cyan-Color".words().camelJoin())
print(".SFNSDisplay".words().camelJoin())
print("Accept-CGU".words().camelJoin())
print("Show-NavCtrl".words().camelJoin())
print("apples.count".words().camelJoin())
print("apples. count".words().camelJoin())
print("multiLine\nKey".words().camelJoin())
```

Gives:
```
CyanColor
SFNSDisplay
AcceptCGU
ShowNavCtrl
Apples.count
MultiLineKey
```

We could argue in considering `x.y` as 2 words too to make that `ApplesCount` instead of having `Apples.count` (that will become `Apples_count` once swiftIdentifier'd) but the rest seems consistent

AliSoftware:
If I use that addition:
```swift
  func punctWords() -> [String] {
    return self.components(separatedBy: .punctuationCharacters).flatMap({ $0.words() })
  }
```

Then we got:
```
"apples.count".punctWords().camelJoin()
// ApplesCount
"a_b.c.d-e_f".punctWords().camelJoin()
// ABCDEF
```

AliSoftware:
After all from a logical standpoint, that's exactly what we want: having a string that is CamelCased, that is upper case at… word boundaries. Be that boundaries be an underscore, a dash, a period, or similar. After all that's how we CamelCase strings in our head, right?

AliSoftware:
You know what? Even `strinng.components(separatedBy: NSCharacterSet.alphanumerics.inverted).camelJoin()` could be sufficient actually, it seems

```swift
func test(strings: [String], apply: (String) -> String) {
  for s in strings {
    let orig = s.replacingOccurrences(of: "\n", with: "\\n").padding(toLength: 20, withPad: " ", startingAt: 0)
    let res = apply(s)
    print("\(orig) --> \(res)")
  }
}

let inputs = ["Cyan-Color", ".SFNSDisplay", "Accept-CGU", "Show-NavCtrl", "apples.count", "apples. count", "multiLine\nKey","foo_bar.baz.qux-yay"]

test(strings: inputs) {
  $0.components(separatedBy: NSCharacterSet.alphanumerics.inverted).camelJoin()
}
/*
Cyan-Color           --> CyanColor
.SFNSDisplay         --> SFNSDisplay
Accept-CGU           --> AcceptCGU
Show-NavCtrl         --> ShowNavCtrl
apples.count         --> ApplesCount
apples. count        --> ApplesCount
multiLine\nKey       --> MultiLineKey
foo_bar.baz.qux-yay  --> FooBarBazQuxYay
*/

**Comment by AliSoftware** [comment]
Ok, continuing the thought, maybe the logical thing to do would be to split using any character that is invalid as a swift identifier character indeed, instead of splitting using any character that is not an alphanumeric. So that we keep any emoji or whatnot untouched.

But then guess what… that would be equivalent of "splitting the string using non-swift-char boundaries, then capitalising the first char of each "word" found, then joining"… which would actually be equivalent of `swiftIdentifier|camelCase|swiftIdentifier` in the end… so maybe you were right all along and I just needed to convince myself… 😇

**Comment by djbe** [comment]
That then breaks things for strings such as `HEADER_TITLE` --> `HEADERTITLE`, and other issues with full uppercase keys.

This works though:
```swift
		string = string
			.components(separatedBy: NSCharacterSet.alphanumerics.inverted)
			.map { titlecase($0) }
			.joined(separator: "_")
		string = try snakeToCamelCase(string, stripLeading: true)
		return StencilSwiftKit.swiftIdentifier(from: string, replaceWithUnderscores: true)
```

djbe:
Oh, you beat me to it, indeed, emoji 🤷‍♂️

djbe:
What we could do, is:
- swiftidentifier
- snakeToCamelCase:"true"
- titlecase
- prefix with `_` if needed

So, move the last 2 items from swiftidentifier into a loose function that we can call separately.

djbe:
Actually, ignore that `titlecase` after the snake fitler, it isn't needed.

AliSoftware:
Ah, looks like a good idea! 👍

Documentation/filters-strings.md
```diff
@@ -166,7 +166,9 @@ This filter accepts a parameter (boolean, default `false`) that controls the pre
 
 ## Filter: `swiftIdentifier`
 
-Transforms an arbitrary string into a valid Swift identifier (using only valid characters for a Swift identifier as defined in the Swift language reference). It will apply the following rules:
+This filter has a couple of modes that you can specifiy using an optional argument (defaults to "normal"):
+
+**normal**: Transforms an arbitrary string into a valid Swift identifier (using only valid characters for a Swift identifier as defined in the Swift language reference). It will apply the following rules:
 
 - Uppercase the first character.
```
AliSoftware: Oh, really, does that filter actually uppercase the first character all the time in our code (even since Swift 3)? Shouldn't by anymore I think… (if it's still the case, might be worth using that PR to make the `normal` mode correct again and stop uppercasing while there's not reason to?)

README.md
```diff
@@ -36,7 +36,7 @@
   * `removeNewlines`: Removes newlines and other whitespace characters, depending on the mode ("all" or "leading").
   * `replace`: Replaces instances of a substring with a new string.
   * `snakeToCamelCase`: Transforms text from snake_case to camelCase. By default it keeps leading underscores, unless a single optional argument is set to "true", "yes" or "1".
-  * `swiftIdentifier`: Transforms an arbitrary string into a valid Swift identifier (using only valid characters for a Swift identifier as defined in the Swift language reference)
+  * `swiftIdentifier`: Transforms an arbitrary string into a valid Swift identifier (using only valid characters for a Swift identifier as defined in the Swift language reference). In "pretty" mode, it will also first apply the snakeToCamelCase filter.
```
AliSoftware: It will not apply the snakeToCamelCase filter first, but last, right?

Sources/SwiftIdentifier.swift
```diff
@@ -78,5 +74,19 @@ func swiftIdentifier(from string: String,
       return ""
     }
   })
-  return prefix + mappedParts.joined(separator: replacement)
+
+  let result = mappedParts.joined(separator: replacement)
+  return prefixWithUnderscoreIfNeeded(string: result, forbiddenChars: exceptions)
+}
+
+func prefixWithUnderscoreIfNeeded(string: String,
```
AliSoftware: I'm not a fan of that function name, because it doesn't tell when it's needed — it doesn't tell that it makes sense only in the context of a Swift identifier.
But I have to admit that I'm not inspired either for a better name, so… ¯\_(ツ)_/¯

Sources/SwiftIdentifier.swift
```diff
@@ -78,5 +74,19 @@ func swiftIdentifier(from string: String,
...
+                                  forbiddenChars exceptions: String = "") -> String {
+
+  let (head, tail) = identifierCharacterSets(exceptions: exceptions)
+
+  let chars = string.unicodeScalars
+  let firstChar = chars[chars.startIndex]
+  let prefix = !head.longCharacterIsMember(firstChar.value) && tail.longCharacterIsMember(firstChar.value) ? "_" : ""
```
AliSoftware: I'm not sure I remember why we have that test logic.
Shouldn't testing against the valid characters for `head` be enough here to determine if we must prefix with `_` or not? If the first character is not in the set of valid head chars, then what's the point of testing it against tail, we will prefix it with `_` anyway…

And then if that tail is indeed not needed, we should probably split the `identifierCharacterSets` function returning a tuple into 2 functions returning a single set each I guess.

Documentation/filters-strings.md
djbe: That... would be a breaking change TBH. And for what? Types will still need an uppercase first letter, so we'll still need to apply a filter.

That's what I meant to say in the discussion in #30, add enum cases for swift versions, and maybe even for the type of identifier (variable, case, enum, struct, etc...). Maybe not so extensive (not so many options), but could be possible.

Although upper first letter is a decent default.

README.md
djbe: Yeah, we changed the order around in the last commit.

Sources/SwiftIdentifier.swift
djbe: That bit of code used to be in the `swiftIdentifier(from:...)` method, and it had to check if it was still a valid character, otherwise that character was going to be replaced by a `_`.

I think (not 100% sure) that the check for tail can be removed here.

Documentation/filters-strings.md
AliSoftware: Well it's debatable. The aim of `swiftIdentifier` is only to make the string valid Swift. Prettifying the string is nice too, but transforming *too much* means that if we don't want that transformation we have to apply the reverse of that transform afterwards to cancel it… And that reverse transform may not always exist.

Just feels odd to over-transform when half of the cases would need it, sure, but the other half doesn't and would need to revert it, so why force to do it in the first place instead of letting it be opt-in is what I mean. The history of this comes back from Swift 2 of course, but we're passed that now.

But indeed that would be a breaking change so maybe let's keep that for a later PR :-/

djbe: The thing is, `snakeToCamelCase` would uppercase the first letter anyway, no way around that.

And I'm not talking about swift 2, but even swift 3/4 types need an uppercase first letter. Only variables and cases don't.

AliSoftware: Right, Right

Issue #58, opened 2017-08-15
Reduce complexity of Strings filters


Antondomashnev:
The file has grown a lot since the addition of new filters by #54. The splitting of the file will improve readability and reduce complexity.

PR #63, opened 2017-08-20
Splitting of Filters+Strings

Antondomashnev:
This resolves the #58, naming selection was pretty difficult, but so far it looks ok to me. @AliSoftware @djbe if you have better names - I'd be happy to change it 😄


Sources/Filters+Strings+Lettercase.swift
```diff
@@ -0,0 +1,157 @@
...
+        return lowercaseCharRegex.firstMatch(in: string, options: .reportCompletion, range: fullRange) != nil
+    }
+
+    /// This returns the string with its first parameter uppercased.
+    /// - note: This is quite similar to `capitalise` except that this filter doesn't
+    ///          lowercase the rest of the string but keeps it untouched.
+    ///
+    /// - Parameter string: The string to titleCase
+    /// - Returns: The string with its first character uppercased, and the rest of the string unchanged.
+    private static func titlecase(_ string: String) -> String {
```
djbe: Rename this to `upperFirstLetter`, `titlecase` is the old name for this filter/function.

Sources/Filters+Strings+Lettercase.swift
```diff
@@ -0,0 +1,157 @@
...
+        return snakeCase
+    }
+
+    /// Converts snake_case to camelCase, stripping prefix underscores if needed
+    ///
+    /// - Parameters:
+    ///   - string: the value to be processed
+    ///   - stripLeading: if false, will preserve leading underscores
+    /// - Returns: the camel case string
+    static func snakeToCamelCase(_ string: String, stripLeading: Bool) throws -> String {
```
djbe: This should probably be moved to private, except that swiftIdentifier depends on it.

Should we have a clearer separation between the Stencil filter functions (that accept and return Any), and the actual filter methods (that accept and return String + typed arguments)?

djbe:
While we're at it, maybe check on codebeat if there are other changes we might want to apply:
https://codebeat.co/projects/github-com-swiftgen-stencilswiftkit-master/report_classes/swift-Filters-Strings
Note that some issues are already resolved by this PR (see the codebeat CI step).

Pods/Pods.xcodeproj/project.pbxproj
```diff
@@ -385,7 +387,9 @@
 				2E92B989035E25169DF9300616F23AAB /* Environment.swift */,
 				0C6D246B203676EC17FA9473D97CB418 /* Filters.swift */,
 				02B0A9A1128124B1A17D5AB1E23A0A45 /* Filters+Numbers.swift */,
-				47DB41142F62E8B5174F3DF281DCD5E3 /* Filters+Strings.swift */,
+				6D1EF02F1F49EC7C00EA91EF /* Filters+Strings+Lettercase.swift */,
+				6D1EF02B1F49E85D00EA91EF /* Filters+Strings+Mutations.swift */,
```
AliSoftware: those short UUIDs intrigue me… is this a new thing in Xcode 9, shorter UUIDs for file references in Xcodeprojs? 🤔

Sources/Filters+Strings+Boolean.swift
```diff
@@ -0,0 +1,54 @@
...
+// StencilSwiftKit
+// Copyright (c) 2017 SwiftGen
+// MIT Licence
+//
+
+import Foundation
+import Stencil
+
+extension Filters {
+  enum Strings {
```
AliSoftware: Mmmh. There's no particular reason why the `enum` would be declared in `+Boolean` and `extension` be used in the other files… But is it worth having a `Filters+Strings.swift` which would only declare an empty `enum Strings{}` just so all `Filters+Strings+Whatever.swift` subsequent file are consistent and all use `extension Filters.Strings`?

Sources/Filters+Strings+Lettercase.swift
AliSoftware: Yeah, maybe internal real functions should be separate to functions having a Filter signature and being used in `registerFilter`… but then where should we put that `static func snakeToCamelCase` method?

Sources/Filters+Strings+Lettercase.swift
AliSoftware: I wonder why that `titlecase` name came back up here, did you copy/paste the code from _before_ the last commits which changed that name when you started working on that split? Or did we miss something when doing our latest merges and let that slip?

AliSoftware:
@Antondomashnev you might want to (1) rebase your branch on top of `master` now that we recently released SwiftGenKit 2.1.0 (as of minutes ago), and also double and triple-check that what you did cut&paste in separate files didn't get changed by recent commits, making you miss those changes (see the `titlecase case`) and risking introducing regressions.

Otherwise I'm ok with the names (or rather I can't think of anything better at the moment :-p), thanks for the PR :)

djbe:
I think the code is based on almost the most recent code. `titlecase` was an old internal method from before the filter got renamed. I think it got missed during the rename and addition of new filters.

Pods/Pods.xcodeproj/project.pbxproj
djbe: It happens when you add the files manually to the pods project. When you run `pod install` you get the longer UUIDs.

Sources/Filters+Strings+Lettercase.swift
Antondomashnev: Regarding the:
>  Should we have a clearer separation between the Stencil filter functions (that accept and return Any), and the actual filter methods (that accept and return String + typed arguments)?

I wanna ask if it makes sense for you to validate `input` not for string, but for string convertable. That would actually be a benefit for Sourcery and simplify the template writing 😄

Sources/Filters+Strings+Lettercase.swift
Antondomashnev: @AliSoftware when I cloned the repo today the initial state was 49af697239

Sources/Filters+Strings+Lettercase.swift
djbe: @Antondomashnev so `let value = value as? StringConvertible, let string = String(describing: value)`?

Pods/Pods.xcodeproj/project.pbxproj
Antondomashnev: Thanks @djbe didn't know it. @AliSoftware I used Xcode 8.x but added the, manually.

Sources/Filters+Strings+Boolean.swift
Antondomashnev: I like the idea, I think it may be better.

Sources/Filters+Strings+Lettercase.swift
djbe: Seems like a good idea to me, we don't care what the data is as long as we can safely convert it to a string.

Antondomashnev: Cool, I guess it worth another PR, as to not mess up things 😄

Sources/Filters+Strings+Lettercase.swift
Antondomashnev: @AliSoftware I think you messed up something as we're still using `titlecase` 😄

Antondomashnev: In c4686c3 it's removed, but I had to create a function `upperFirstLetter` that doesn't throw and takes `String` parameter to be able to use in:
```
let comps = string.components(separatedBy: "_")
unprefixed = comps.map { upperFirstLetter($0) }.joined(separator: "")
```

Sources/Filters+Strings+Boolean.swift
Antondomashnev: I've done it in 73ca5da. But there is one thing worth to mention. The order of compile sources must be like that, so `Filters+Strings.swift` is compiled first 😄

Sources/Filters+Strings+Lettercase.swift
Antondomashnev: In c4686c3 it's removed, but I had to create a function `upperFirstLetter` that doesn't throw and takes `String` parameter to be able to use in:
```
let comps = string.components(separatedBy: "_")
unprefixed = comps.map { upperFirstLetter($0) }.joined(separator: "")
```

Sources/Filters+Strings+Boolean.swift
Antondomashnev: I've done it in 73ca5da. But there is one thing worth to mention. The order of compile sources must be like that, so `Filters+Strings.swift` is compiled first 😄

Antondomashnev:
@AliSoftware @djbe the build is failing because the file `Filters+Strings` has not been compiled yet, when the `Filters+Strings+...` are compiling. I've worked around it with `build` and `test` from Xcode, but haven't found solution yet for SPM. If you have an idea that would be helpful.

The simplest way to fix it would be to keep the decalaration of `enum Strings` as it was in `Filters+Strings+Boolean` but I do like much more the separate file with declaration.

AliSoftware:
I'm not sure I understand. How come in the case where the `enum` is in `Filters+Strings` it might not be compiled before, but when it's in `Filters+Strings+Boolean` it's guaranteed to be compiled before `Filters+Strings+Lettercase` for example?

djbe:
@Antondomashnev Is it because you added the files to the pods project manually? Order really shouldn't matter. Try running `bundle exec pod install`.

Antondomashnev:
@AliSoftware I don't understand as well. 

But as I wrote in the comment above:
> The order of compile sources must be like that, so `Filters+Strings.swift` is compiled first 😄

Antondomashnev:
@djbe tried it, didn't help. Also tried to remove references for these files, add them again and `bundle exec pod install`, still no luck.

Antondomashnev:
@djbe @AliSoftware order definitely doesn't matter, but SPM still fails 😄

AliSoftware:
Does SPM uses -WMO? Or if not any way to enable it? I wonder if that would make a difference, forcing to compile it as a whole…

Antondomashnev:
@AliSoftware do you mean whole module optimization?

AliSoftware:
Yup

Antondomashnev:
@AliSoftware seems it does use it in release build and `swift build -c release` also doesn't work:
> Also the Swift Package Manager compiles with whole-module optimizations in release builds.

Antondomashnev:
Yeah, I've made it working. So it seems for the SPM the file names are still matter and I think it's somehow sorted so the `Filters+Strings.swift` goes after `Filter+Strings+....swift`.
I've renamed `Filters+Strings.swift` to `Filters+Strings+Base.swift` so if we sort it alphabetically which is a guess from my side it becomes the first through all of extensions.

AliSoftware:
Ok, that's cool… but it's still worth a JIRA in bugs.swift.org to me though, as I think that shouldn't be the case 😉

Antondomashnev:
@AliSoftware very valid point, create [one](https://bugs.swift.org/browse/SR-5734). Btw I still can not merge the PRs, maybe `master` is protected with some rules.

AliSoftware:
Yup, only the CoreTeam have right to merge into master :wink:

PR #65, opened 2017-08-28
Parse string function

### LINKED ISSUE #64: Allow to use strings filters on CustomStringConvertable
This was proposed in #58. The idea is that at the moment all strings filters from `Filters+Strings...` expect `String` parameter and throw exception if it's not satisfied, however we could change this rule to accept all parameters that can be described as a string, which `CustomStringConvertable` is made for 😄 

We'll also make use of it in [Sourcery](https://github.com/krzysztofzablocki/Sourcery)

Antondomashnev:
This resolves #64, but with one tweak, I've realised that `CustomStringConvertible` is too broad protocol, that has many adoptions which not really a fit to be a filter parameter, so I used `LosslessStringConvertible`. I think it's a good tradeoff between making input not that strict, but still with some sense.

Regarding the functions I've tried to follow the pattern in `Filters.swift` but I had to introduce two functions - parse array of `Any?` and just `Any?` as it reduced the possible boilerplate code. I've also tried to use the same name for these two functions:
```
func parseString(from value: Any?) throws -> String
func parseString(from arguments: [Any?], at index: Int = 0) throws -> String
```
but it lead to confusion for compiler as `Any?` could be anything, also an array of `Any?` 😄 

I thought maybe we can have an `Extractor` instead of having these functions in `Filters`, what do you think?


CHANGELOG.md
```diff
@@ -14,7 +14,9 @@ _None_
 
 ### New Features
 
-_None_
+* Accept `LosslessStringConvertible` input for strings filters.  
+  [Antondomashnev](https://github.com/antondomashnev)
+  [#59](https://github.com/SwiftGen/StencilSwiftKit/pull/63)
```
djbe: Please correct this.

Sources/Filters.swift
```diff
@@ -13,6 +13,33 @@ enum Filters {
...
+  }
+
+  /// Parses filter arguments for a string value, where accepted objects must conform to 
+  /// `CustomStringConvertible`
+  ///
+  /// - Parameters:
+  ///   - arguments: an array of argument values, may be empty
+  ///   - index: the index in the arguments array
+  /// - Throws: Filters.Error.invalidInputType
+  static func parseStringArgument(from arguments: [Any?], at index: Int = 0) throws -> String {
```
djbe: To keep the naming consistent with the other filters, we could consider making the index mandatory (no default value). And modify the other parse argument functions (bool/enum) to require an index as well.

Antondomashnev: Honestly I don't see the benefit for this, we'll increase the number of repeatable code in order to have naming consistency. I think it doesn't worth it 😄

CHANGELOG.md
Antondomashnev: Thanks, I've also corrected previous entry, I think we missed it during the review of #63.
Check ad8c792

Antondomashnev: Thanks, I've also corrected previous entry, I think we missed it during the review of #63.
Check ad8c792

Antondomashnev:
@djbe ping 😄

PR #69, opened 2017-10-05
Additional boolean templates;

Antondomashnev:
This is a result of the work on the Sourcery [task](https://github.com/krzysztofzablocki/Sourcery/issues/337). It brings the negative counterpart for `Strings+Boolean` filters:
* `!contains`
* `!hasPrefix`
* `!hasSuffix`

@AliSoftware @djbe I'm not sure whether we need to update docs or not, from my point of view we don't 😄

### DIFF
--- a/Sources/Environment.swift
+++ b/Sources/Environment.swift
@@ -15,6 +15,12 @@ public extension Extension {
 
   // MARK: - Private
 
+  private func registerFilter(_ name: String, filter: @escaping Filters.BooleanWithArguments) {
+    typealias GenericFilter = (Any?, [Any?]) throws -> Any?
+    registerFilter(name, filter: filter as GenericFilter)
+    registerFilter("!\(name)", filter: { value, arguments in try !filter(value, arguments)} as GenericFilter)
+  }
+
   private func registerNumbersFilters() {
     registerFilter("hexToInt", filter: Filters.Numbers.hexToInt)
     registerFilter("int255toFloat", filter: Filters.Numbers.int255toFloat)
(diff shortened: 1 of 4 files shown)


djbe:
Hmmm, why use these `!stuff` instead of `not stuff`? Conciseness?

Do we know if Kyle is planning on adding parsing for a shorthand `not`?

Antondomashnev:
@djbe I didn't ask @kylef, maybe it would be a good addition for the Stencil itself.

AliSoftware:
I think we should wait for @kylef to chime on this one. Not sure adding a `!` counterpart instead of just using `not` is really worth it, especially if it's only applied to some filters (the 3 string boolean filters StencilSwiftKit adds here) and not all (all boolean filters Stencil proper declares)

kylef:
I am wondering what the use-case is here. Would they be used in an `if` statement? Which there is already a `not`.

Thanks to https://github.com/kylef/Stencil/pull/143 there is an `in` expression:

```html+django
{% if "thing" not in collection %}
{% endif %}
```

(I don't recall if filters are supported in if statements, but something like the following *may* work or be easy to implement):

```html+django
{% if not collection|hasPrefix:"test" %}
{% endif %}
```

Antondomashnev:
@kylef @AliSoftware to be honest, the most reason is to be able to easily remove filters from `Sourcery` that are now in `StencilSwiftKit`. 
The link to these three filters is [here](https://github.com/krzysztofzablocki/Sourcery/blob/master/Sourcery/Generating/Template/Stencil/StencilTemplate.swift#L122) and as you may find the `registerBoolFilterOrWithArguments ` generates also `!{filter}`.

krzysztofzablocki:
`not` operator does not allow for chaining e.g.
`{% for type in types.all|!inheriting.BaseClass|!inheriting.BaseProtocol %}`

AliSoftware:
Again, my only concern is that some boolean filters (the ones in Stencil proper) will not have this kind of `!` operators while the ones declared in StencilSwiftKit will.

I think we can approve this for now, at least to go forward, but it would be better to convince @kylef to include that logic in Stencil proper in a later time for consistency between all boolean filters.

Sources/Environment.swift
```diff
@@ -15,6 +15,11 @@ public extension Extension {
 
   // MARK: - Private
 
+  private func registerBooleanFilterWithArguments(_ name: String, filter: @escaping Filters.BooleanWithArguments) {
```
AliSoftware: Just nitpicking, but I wonder if we shouldn't just call this method `registerFilter` like all the others — and let the type overloading do the magic of using the proper version of `registerFilter` depending on the kind of filter passed…

krzysztofzablocki: @Antondomashnev up to you

Antondomashnev: Hah, I've tried it but I forgot that call original `registerFilter` inside the `registerBooleanFilterWithArguments ` and it leads to the recursion 😞 @krzysztofzablocki @AliSoftware do you have any idea how to workaround it?

AliSoftware: Using `as` to specify the overload you want to call should do it:

```swift
typealias AnyFilter = (Any?, [Any?]) throws -> Any?
registerFilter(name, filter: filter as AnyFilter)
registerFilter("!\(name)", filter: { value, arguments in try !filter(value, arguments) } as AnyFilter)
```

PR #73, opened 2017-12-22
Stencil update

ilyapuchka:
There were few nice additions and fixes in latest releases of Stencil so would be cool to bump it.


AliSoftware:
Not sure why CI fails on SPM, while you seem to have updated the Package.swift appropriately?

ilyapuchka:
worked after updated package pins

PR #78, opened 2018-03-27
Use SwiftLint via CocoaPods

djbe:
Same as https://github.com/SwiftGen/SwiftGen/pull/401


.swiftlint.yml
```diff
@@ -0,0 +1,14 @@
+disabled_rules:
+ - type_name
+
+opt_in_rules:
+ - force_unwrapping
```
AliSoftware: 👍, + 💯

Tests/StencilSwiftKitTests/ParseBoolTests.swift
```diff
@@ -8,17 +8,20 @@ import XCTest
 @testable import StencilSwiftKit
 
 class ParseBoolTests: XCTestCase {
-
   func testParseBool_TrueWithString() throws {
+    // swiftlint:disable force_unwrapping
     XCTAssertTrue(try Filters.parseBool(from: ["true"])!)
     XCTAssertTrue(try Filters.parseBool(from: ["yes"])!)
     XCTAssertTrue(try Filters.parseBool(from: ["1"])!)
+    // swiftlint:enable force_unwrapping
```
AliSoftware: Maybe we should instead use `XCTAssertTrue((try …) ?? false)` or `XCTAssertEquals(try …, true)` to avoid the force-unwrap here? Especially, in case there's a regression one day and the test  `parseBool(from: ["true"])` starts to fail and return an unexpected `nil`, that force-unwrap will make the test harness crash and stop the whole test process, instead of marking that test as failed and continue with other test cases…

circle.yml
```diff
@@ -4,7 +4,6 @@ machine:
 
 dependencies:
   post:
-    - bundle exec rake lint:install
```
AliSoftware: Same questions as in https://github.com/SwiftGen/SwiftGen/pull/401#discussion_r177748654

rakelib/lint.rake
```diff
@@ -1,62 +1,31 @@
...
-      Utils.print_header 'Linting the code'
-      config = Pathname.getwd + '.swiftlint.yml'
-      config = config.exist? ? "--config #{config}" : ''
-      Utils.run(%(swiftlint lint --no-cache --strict --path Sources #{config}), task)
-    end
+  desc 'Lint the code'
+  task :code do |task|
+    Utils.print_header 'Linting the code'
+    config = Pathname.getwd + '.swiftlint.yml'
+    Utils.run(%(#{SWIFTLINT} lint --strict --path Sources --config "#{config}"), task)
```
AliSoftware: Same q as in https://github.com/SwiftGen/SwiftGen/pull/401#discussion_r177749127

Tests/StencilSwiftKitTests/ParseBoolTests.swift
djbe: We'd have to be careful then to provide the "inverse" value as a fallback for each test. Here it's true, for the `parseBool(from: ["false"])` it should be false, ...

circle.yml
djbe: Same answer :trollface:

Tests/StencilSwiftKitTests/ParseBoolTests.swift
AliSoftware: Yeah that's why I'd prefer to use the more readable `XCTAssertEquals(…, true)` there rather than `XCTAssertTrue(… ?? false)`

Scripts/SwiftLint.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+SWIFTLINT="$(PWD)/Pods/SwiftLint/swiftlint"
+CONFIG="$(PWD)/.swiftlint.yml"
```
AliSoftware: Same as https://github.com/SwiftGen/SwiftGen/pull/401#discussion_r180075706

PR #80, opened 2018-03-27
Migrate to Swift 4

djbe:
As is, this PR won't support swift 3 because of the `.characters` change (in `Sources/Filters+Strings.swift`).

We have a couple of options (some can be combined):
- Keep the old `.characters` code. This gives warnings in Swift 4.
- Drop support for Swift 3. This might be a breaking change for current users.
- Surround with `#if swift(>=4.0)` and have 2 code paths.
- Add `swift_version` attribute to the podspec. This is a breaking change for older CocoaPods versions, unless we maybe also add `cocoapods_version`?

My preference goes to dropping Swift 3 support, and adding the attributes to the podspec. The only disadvantage might then be for users of Xcode versions older than 9. Those users can always stick with an older version of StencilSwiftKit?


djbe:
Note: the multiline strings change is also unsupported in Xcode 8. That needs Swift 3.2.

AliSoftware:
Summoning @ilyapuchka & @krzysztofzablocki as they are the main users of that framework in Sourcery, aside from us in SwiftGen.

I personally think at that point we can migrate to Swift 4, and Swift 3 users could just keep using old versions of StencilSwiftKit if they're not ready to migrate. But in the meantime, iirc Stencil itself is still in Swift 3 and hasn't migrated yet?

djbe:
Checking Sourcery, they're on Swift 4:
https://github.com/krzysztofzablocki/Sourcery/blob/master/CHANGELOG.md#internal-changes

djbe:
If we add the attributes to the podspec, it won't matter what swift version the client project is, as cocoapods'll set the swift version for only our pod.

AliSoftware:
Ah, yes, of course. (That sweet feature of Swift 4 migration not forcing every other frameworks to be Swift 4 👌 )

ilyapuchka:
Yeah, we are fine, just make sure that it still works with SPM as there are some issues with Commander-PathKit-Stencil combination because Stencil is still Swift3.

djbe:
Right, these last changes should cover what was discussed.

I'll leave the new Stencil version (and CocoaPods) for another PR.

AliSoftware:
Do we have a check done by the CI about trying to compile with SPM, to ensure that we don't forget to update the Package.swift & al? I see you did it in this PR, wondering if you just happen to not forget it and think about it by yourself or if it was caught by the CI

PR #87, opened 2018-05-10
Advanced Set tag

djbe:
Refs https://github.com/SwiftGen/SwiftGen/pull/379#issuecomment-388035669.

This feature adds a variation of the `set` tag to directly set the value of a variable to the contents of another expression, without rendering it down to a `String` (like the current `set` does).

Users will be able to use the `set` tag in 2 ways:
- `{% set myVar %}...{% endset %}` to render and store everything between the start and end tag into the variable.
- `{% set myVar someOtherVar.prop1.prop2 #}` to evaluate and store an expression's result into the variable.

This is particularly useful if the expression is a complex object such as an `Array` or a `Dictionary`.

### DIFF
--- a/Documentation/tag-set.md
+++ b/Documentation/tag-set.md
@@ -7,7 +7,7 @@ This tag stores a value into a variable for later use.
 | Name      | Description                                                            |
 |-----------|------------------------------------------------------------------------|
 | Tag Name  | `set`                                                                  |
-| End Tag   | `endset`  or N/A (if you're creating an alias of an existing variable) |
+| End Tag   | `endset` or N/A (if you're creating an alias of an existing variable) |
 | Rendering | Immediately; no output                                                 |
 
 | Parameter  | Description                                 | 
@@ -18,7 +18,7 @@ This tag stores a value into a variable for later use.
 The parameters and tags of this node depend on which mode you want to use:
 
 - Use `{% set myVar %}...{% endset %}` to render and store everything between the start and end tag into the variable.
-- Use `{% set myVar someOtherVar.prop1.prop2 #}` to evaluate and store an expression's result into the variable.
+- Use `{% set myVar someOtherVar.prop1.prop2 %}` to evaluate and store an expression's result into the variable.
 
 _Example of render:_ `{% set myVar %}hello{% endset %}`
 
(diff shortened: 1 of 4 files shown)


Documentation/tag-set.md
```diff
@@ -4,30 +4,42 @@ This tag stores a value into a variable for later use.
...
+| Parameter  | Description                                 | 
+|------------|---------------------------------------------|
+| Name       | The name of the variable you want to store. |
+| Expression | (Optional) The value to be stored.          |
 
-_Example:_ `{% set myVar %}hello{% endset %}`
+The parameters and tags of this node depend on which mode you want to use:
+
+- Use `{% set myVar %}...{% endset %}` to render and store everything between the start and end tag into the variable.
+- Use `{% set myVar someOtherVar.prop1.prop2 #}` to evaluate and store an expression's result into the variable.
```
AliSoftware: I've seen you twice (here and in the PR description) tell that this ends with `#}` not `%}`. Is this intentional or a double typo?

Documentation/tag-set.md
```diff
@@ -4,30 +4,42 @@ This tag stores a value into a variable for later use.
...
 
-| Name      | Description                      |
-|-----------|----------------------------------|
-| Tag Name  | `set`                            |
-| End Tag   | `endset`                         |
-| Rendering | Immediately; no output           |
+| Name      | Description                                                            |
+|-----------|------------------------------------------------------------------------|
+| Tag Name  | `set`                                                                  |
+| End Tag   | `endset`  or N/A (if you're creating an alias of an existing variable) |
```
AliSoftware: Nitpick: one space too many before 'or'

Sources/SetNode.swift
```diff
@@ -7,33 +7,54 @@
 import Stencil
 
 class SetNode: NodeType {
+  enum Content {
+    case nodes([NodeType])
+    case value(Resolvable)
+  }
```
AliSoftware: Damn this quite confused me… `nodes` is actually used when rendering to a single string value, and `value` is used when pointing to a node tree/object… 😅

I now get the logic but I needed to read the code twice to understand and be sure you didn't invert the logic but my reading did ^^

Maybe renaming `case value` to `case reference(to:)` would have helped? Idk

Documentation/tag-set.md
djbe: I have no idea how that got in there 😆 It's in the PR description because I copy/pasted it.

PR #90, opened 2018-07-03
Linux testing on CI

janderit:
Note: This PR originally fixed some bugs with NSString on Linux. That has been fixed in #94. Now this PR is for testing SPM support on Linux.

StencilSwiftKit was not usable building on Linux due to two string to NSString coercions. The Linux swift compiler does not recognize the coercion.

Replacing them with NSString(string: ...) allows building projects using StencilSwiftKit on Linux.


djbe:
Totally agree with @AliSoftware, if we add Linux support we should test it on CI.

djbe:
Yeah so, `ruby` is missing from that docker image (for linux), so no `gem` or `bundler` commands... Anyone have a docker image handy that fits our requirements, or wants to create one?

djbe:
Right, so I've finally got linux tests working on CI. We should update the job names GitHub expects after the merge.

Note that this introduces `LinuxMain.swift` and `XCTestManifests.swift`, both files are autogenerated by `swift test --generate-linuxmain`. We can keep using this (and should add a Rake task for it), or use Sourcery to generate the needed code.

djbe:
Oh hey, forgot about this, we merged #94 that fixes a similar issue. I'll rebase this PR to just add the linux support part.

djbe:
@RahulKatariya With what Swift version did you run the tests on Linux?

There's a bug (https://github.com/apple/swift-corelibs-foundation/pull/1536) that causes our Linux tests to crash when using `NSString.standardizingPath` (used by `Path.normalize()`). It's been fixed, but no idea in which Swift version (we're using 4.1.x).

djbe:
Checked and the fix is neither in the swift 4.1 or 4.2 branches, only in master. @AliSoftware do you think they'd be open to cherry-picking that change into those branches, and would we go about it? Comment on that PR? Open an issue?

AliSoftware:
For reference, we've pinged some Swift core team members, and @millenomi told us they'll look into merging the fix soon, but that it will not be included in the upcoming release of Swift anyway (too late already). So we'll have to rely on a workaround (at least in the test suite, if a local fix is possible without requiring a dedicated version of `PathKit` just for that workaround, preferably)

RahulKatariya:
@djbe The path methods were causing the compilation error on Linux, I didn't really use those.

djbe:
@AliSoftware I'm merging this now as it's "fixed", we can revisit this later if we want, but at least we won't block the PRs in SwiftGen anymore.

AliSoftware:
I'm ok with merging to avoid waiting.

A comment in the code referencing 
https://github.com/apple/swift-corelibs-foundation/pull/1536 would have been nice though, to understand why we do that `#if os(Linux)` workaround and to be able to quickly check when it's possible to remove the workaround in the future.

PR #93, opened 2018-07-21
Fix Xcode 10b / Swift 4.2 compiler issue (type-checker timeout)

### LINKED ISSUE #92: Build failure using Swift 4.2 from Xcode 10 Beta
I'm running into an issue building StencilSwiftKit with SPM using Swift 4.2.

```
/.build/checkouts/StencilSwiftKit.git-5443537928707802040/Sources/SwiftIdentifier.swift:16:55: error: the compiler is unable to type-check this expression in reasonable time; try breaking up the expression into distinct sub-expressions
private let headRanges: [CountableClosedRange<Int>] = [
```

It looks like the same issue was detected by the Swift source compatibility suite: https://bugs.swift.org/browse/SR-8046

Has anyone else seen this / know of a work around?

AliSoftware:
This fixes #92 by adding hints to help the type-checker and avoid a timeout during compilation.


AliSoftware:
I'd say the change doesn't harm much (more verbose sure but ¯\\_(ツ)_/¯ ) so why not

djbe:
Works for me. We can check later and remove the hints, although I'd keep the typealias, that's nice.

PR #95, opened 2018-08-26
Stencil 0.12

djbe:
Updated our code as needed:
- Add `token` attribute for better errors support
- Switched to resolvable where possible 
- Had to make the `CallableBlock` a non-`NodeType` to avoid rendering during (variable) resolvable resolution.


Sources/CallMacroNodes.swift
```diff
@@ -6,10 +6,15 @@
...
 import Stencil
 
-struct CallableBlock: NodeType {
+struct CallableBlock {
   let parameters: [String]
   let nodes: [NodeType]
 
+  init(parameters: [String], nodes: [NodeType], token: Token? = nil) {
+    self.parameters = parameters
+    self.nodes = nodes
```
AliSoftware: Not sure I get the use of the `token` parameter here given it's unused?

Sources/MapNode.swift
```diff
@@ -8,28 +8,33 @@
...
   let nodes: [NodeType]
+  let token: Token?
 
   class func parse(parser: TokenParser, token: Token) throws -> NodeType {
     let components = token.components()
 
-    guard components.count == 4 && components[2] == "into" ||
-      components.count == 6 && components[2] == "into" && components[4] == "using" else {
+    func hasToken(_ token: String, at index: Int) -> Bool {
+      return components.count > (index + 1) && components[index] == token
```
AliSoftware: I nitpick but I feel it's more readable to write it as `index < components.count - 1` instead of the other way around

Sources/CallMacroNodes.swift
djbe: It's mandatory for the new error functionality (https://github.com/stencilproject/Stencil/pull/167).

Sources/MapNode.swift
djbe: Yeah I just copied this from stencil (seemed clearer instead of checking a whole bunch of indices). Will update this.

AliSoftware: Another way is to write `collection.indices ~= index` (or is it the other way around I always forget)

djbe: That's the first time I see that (the match operator) used for checking bounds, interesting.

AliSoftware: Well I haven't checked if that works on indices (which is what? A collection?), I know it works on Ranges but didn't check in code that it works work with indices… so curious if that's actually possible :wink:

PR #105, opened 2018-10-01
Fix `swiftIdentifier` crash with empty strings

djbe:
Fixes https://github.com/SwiftGen/SwiftGen/issues/528.

### DIFF
--- a/Sources/StencilSwiftKit/SwiftIdentifier.swift
+++ b/Sources/StencilSwiftKit/SwiftIdentifier.swift
@@ -114,10 +114,9 @@ enum SwiftIdentifier {
   static func prefixWithUnderscoreIfNeeded(string: String,
                                            forbiddenChars exceptions: String = "") -> String {
 
-    let (head, _) = identifierCharacterSets(exceptions: exceptions)
+    guard let firstChar = string.unicodeScalars.first else { return "" }
 
-    let chars = string.unicodeScalars
-    let firstChar = chars[chars.startIndex]
+    let (head, _) = identifierCharacterSets(exceptions: exceptions)
     let prefix = !head.longCharacterIsMember(firstChar.value) ? "_" : ""
 
     return prefix + string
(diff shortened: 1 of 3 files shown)


Sources/StencilSwiftKit/SwiftIdentifier.swift
```diff
@@ -114,10 +114,10 @@ enum SwiftIdentifier {
   static func prefixWithUnderscoreIfNeeded(string: String,
                                            forbiddenChars exceptions: String = "") -> String {
 
-    let (head, _) = identifierCharacterSets(exceptions: exceptions)
+    guard let firstChar = string.unicodeScalars.first,
+      !string.isEmpty else { return "" }
```
AliSoftware: `string.isEmpty` can't ever be false if `string.unicodeScalars.first` is not `nil`, right? If `string` is empty, then `string.unocodeScalars.first` will return `nil` and already make the condition fail anyway

djbe: You'd guess so, wouldn't you? Give it a try, for some absolutely bizarre reason, the `isEmpty` check is needed 🤷‍♂️

AliSoftware: Wait seriously?

isEmpty checks on characters, unicodeScalars is different, so that could make sense, maybe… but can't see an example

Do you mean that if we don't test isEmpty, the test on "" fails?

djbe: It throws an exception (SIGINT or something I think)

AliSoftware: Just tested by cloning your branch and removing that `!string.isEmpty` in the condition and all test passed successfully, without SIGINT?

djbe: Which Xcode version?

AliSoftware: 10.0 final

PR #111, opened 2019-01-08
`import` tag for reusable macros

### LINKED ISSUE #91: reusable macro templates
I've a lot of similar templates with a lot of common macros, I'd like to just have a copy of each single macro stored inside a common file included by all templates. Unfortunately it looks like I can't just {% include .. %} a macro template and use the macros defined there. When I do call them they are not defined. Is it possible to export macros in a way that could be used from the "outer" context ? thanks in advance

cornr:
Implemented @nferruzzi node to import other Templates

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,9 @@
 * Updated to Stencil 0.15.0, which contains many improvements, chief amongst them is whitespace behaviour control.  
   [David Jennes](https://github.com/djbe)
   [#159](https://github.com/SwiftGen/StencilSwiftKit/pull/159)
+* Added an `import` tag for reusing macro's in multiple templates from a common imported file.  
+  [David Jennes](https://github.com/djbe)
+  [#111](https://github.com/SwiftGen/StencilSwiftKit/pull/111)
 
 ### Bug Fixes
 
--- a/Documentation/tag-import.md
+++ b/Documentation/tag-import.md
@@ -1,0 +1,48 @@
+# Tag: "Import"
+
+This tag loads and executes another template, but will **never** render any output. It's only purpose is for loading [macros](tag-macro.md) and [setting some variables](tag-set.md).
+
+## Node Information
+
+| Name      | Description            |
+|-----------|------------------------|
+| Tag Name  | `import`               |
+| End Tag   | N/A                    |
+| Rendering | Immediately; no output |
+
+| Parameter  | Description                                | 
+|------------|--------------------------------------------|
+| Template   | The name of the template you want to load. |
+
+The template name will depend on how your environment's loader is configured:
+
+- If you're using a file loader, template should be a valid file name.
+- If you're using a dictionary loader, template should be a key in that dictionary.
+
+## When to use it
+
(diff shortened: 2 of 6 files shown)


cornr:
I'm not sure with xcodeproj changes.
To bootstrap the project and run the tests I did `swift package generate-xcodeproj`

djbe:
I don't think we've ever used `swift package generate-xcodeproj`. If you add files you should run `pod install`.

djbe:
It might be better to undo the changes from `swift package generate-xcodeproj`, it seems to have added some unneeded files.

cornr:
@djbe thanks for your help. I reverted the changes from `swift package generate-xcodeproj` and executed `bundle exec pod install` and now I am running into https://github.com/CocoaPods/CocoaPods/issues/7708

nferruzzi:
thank you guys! I totally forgot about this <3

cornr:
Sorry the PR is not working as expected. While I was testing the import Tag with Sourcery I figured that the loaded template is not populated to the context and therefore the macros are not visible.
@nferruzzi in your code snippet you call `_ = try template.render(context)` which is a non open function and therefore not visible in the ImportNode. The open `template.render(context.flatten())` function uses a new context.
Any ideas to populate the new parsed tokens into the current context?

nferruzzi:
sorry but I think I may not be of great help now, too many months since I created that node, but I keep using it daily and without any problem so it should be something minor; if I do remember properly the trick `_ = ..` was to avoid populating the output with the imported file output and `template.render(context.flatten())` was to inject the context with the macro content. 

My usage, is like this :
```
{% import "Label.swift.stencil" %}
{% import "Button.swift.stencil" %}
{% import "Activity.swift.stencil" %}

... from here I can {%call anyMacro %}
```

djbe:
@cornr I've made a separate PR that updates cocoapods & swiftlint (#112), which has been merged. You should rebase this PR on master to get those changes.

Do note that if you run `pod install`, make sure to afterwards (in Xcode) select the `Pods.xcodeproj` and set the text indentation settings back to indent with `spaces` and `2`:
screenshot 2019-01-08 at 23 33 33
This ensures everyone is writing with the same text editor settings.

cornr:
@nferruzzi your code does not work with the current Stencil Release as its using a non open function: `template.render(context)`

@djbe I wonder why the usage of the `include`-tag does not work with macros?
I checked the code [Include.swift](https://github.com/stencilproject/Stencil/blob/master/Sources/Include.swift) and it looks pretty much the same as @nferruzzi implementation. Maybe the StencilSwiftKit tags are not registered in the used `Environment`.

djbe:
The difference is subtle, but quite important:
- the `include` tag from Stencil surrounds the `render` call with a `context.push(dictionary: subContext) {`
- whereas this `import` tag does not

Surrounding the render with a context push causes all changes during that render (such as creating macro's, setting variables, etc…) to only be applied temporarily, and "popped" away when the render is done.

djbe:
The current implementation was broken, because it was applying the changes to `context.flatten()`, which generates a dictionary **copy**. So the original context is not modified as it should be.

I've updated the implementation and written tests that actually test modification. The implementation currently does **not** compile, because it depends on the non-open function `render` which accepts a context. We'll need to modify Stencil to open this up first.

PR #112, opened 2019-01-08
Update CocoaPods & SwiftLint

djbe:
CocoaPods 1.6 beta has been quite stable (I've used in many many production projects).

I've also enabled some extra SwiftLint rules:
- the only one I couldn't enable is `xct_specific_matcher`, as it fails on https://github.com/SwiftGen/StencilSwiftKit/blob/master/Tests/StencilSwiftKitTests/ParseBoolTests.swift#L12 where we check optional booleans.
- `multiline_arguments_brackets` is a good rule, but it annoyingly triggers on multiline strings.


Tests/StencilSwiftKitTests/CallNodeTests.swift
```diff
@@ -120,20 +129,26 @@ class CallNodeTests: XCTestCase {
 
     // not enough arguments
```
AliSoftware: Strange indentation of comments here and below too 🤔

PR #119, opened 2019-03-03
Swift 5

djbe:
Few things:
- `lowerFirstWord` crashes in swift 5 because `scalars.index(after: start)` crashes for an empty string. I've added a check for `isEmpty` before it, or we could invert the "after start" and "before end" checks. TBH, I think the `isEmpty` check is clearer in its intent.
- We don't have a docker image with Swift 5 support yet.
- Maybe we should drop swift versions older than 4.2.


.circleci/config.yml
```diff
@@ -80,8 +78,40 @@ jobs:
...
+      - run:
+          name: Run all tests
+          command: bundle exec rake spm:test
+      - *store-artifacts
+
+  spm-linux-4.1-build-and-test:
     <<: *linux-config
+    docker:
+      - image: norionomura/jazzy:swift-4.1
+    steps:
```
AliSoftware, line 100: You could create a YAML alias on line 85 (`&spm-steps`) and reference it here (`*spm-steps`) instead of duplicating them

.circleci/config.yml
```diff
@@ -80,8 +78,40 @@ jobs:
...
+      - run:
+          name: Run all tests
+          command: bundle exec rake spm:test
+      - *store-artifacts
+
+  spm-linux-4.2-build-and-test:
+    <<: *linux-config
+    docker:
+      - image: norionomura/jazzy:swift-4.2
     steps:
```
AliSoftware, line 107: And `*spm-steps` here too then

.ruby-version
```diff
@@ -1 +1 @@
-2.3.7
+2.4.5
```
AliSoftware, line 1: If there's no need for newest additions in latest Ruby versions we should try to keep the Ruby version shipped with latest version of macOS (don't even know which it is and if 2.3.7 was already that) so that people not wanting to use rvm/rbenv won't have to worry. Except if you saw features in more recent Ruby versions that we needed?

Sources/StencilSwiftKit/Filters+Strings.swift
```diff
@@ -95,7 +95,8 @@ extension Filters.Strings {
     while idx < scalars.endIndex, let scalar = UnicodeScalar(scalars[idx].value), characterSet.contains(scalar) {
       idx = scalars.index(after: idx)
     }
-    if idx > scalars.index(after: start) && idx < scalars.endIndex,
+    if !string.isEmpty,
```
AliSoftware, line 98: 👍

.ruby-version
djbe: I actually tried that in this PR (and the other PR in SwiftGen), but it starts complaining about an unknown version of ruby, so 🤷‍♂️

.circleci/config.yml
djbe: Right.

djbe:
I'm curious as to why linux swift 4.0 can't find Stencil... It has a `Package.swift` with the tools version set to 4.0. https://github.com/stencilproject/Stencil/blob/master/Package.swift

.circleci/config.yml
```diff
@@ -6,15 +6,13 @@ defaults:
...
-      xcode: "10.0.0"
+      xcode: "10.2.0"
     shell: /bin/bash --login -eo pipefail
   - &linux-config
     environment:
       CIRCLE_ARTIFACTS: /tmp/circleci-artifacts
       CIRCLE_TEST_REPORTS: /tmp/circleci-test-results
       BUNDLE_PATH: vendor/bundle
-    docker:
-      - image: norionomura/jazzy:swift-4.2.0
```
AliSoftware, line 17: > We don't have a docker image with Swift 5 support yet.

Those Docker images with Swift 5 are now since available: https://github.com/norio-nomura/docker-jazzy

djbe:
This PR is superseded by #127, which (amongst other things) upgrades to Swift 5.

PR #122, opened 2019-06-10
Adds Support for Swift 5

hectormatos2011:
Hello!

# Internal Change
It's been bothering me that when I use Swift 5 with StencilSwiftKit, a few warnings pop up due to outdated usages of older Swift APIs. By updating this library to support Swift 5, those warnings will go away. I also fixed the warning for redundant usages of `public` within `public` extensions.

Since [Stencil](https://github.com/stencilproject/Stencil) no longer supports versions of [Swift earlier than 4.1](https://github.com/stencilproject/Stencil/pull/228), I updated the library to do the same by dropping support as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,12 @@ _None_
   [David Jennes](https://github.com/djbe)
   [#116](https://github.com/SwiftGen/StencilSwiftKit/pull/116)
 
+## 2.8.0
+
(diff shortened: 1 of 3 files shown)


djbe:
Thank you for your contribution.

I know it's been a while, and in the mean time, this has been superseded by #127, which (amongst other things) adds support for Swift 5.

PR #123, opened 2020-01-09
Proposal: Support replacing regular expressions

JanGorman:
This is a proposal to add support for replacing regular expressions instead of just substrings. I've added this to the regular replace function. This would require you to wrap your regular expression in `/{regex goes here}/`.
A more explicit alternative would be of course to support a new `replaceRegex:"{regex}","{replacement}"` syntax on top of the current `replace`.

What are your thoughts?


djbe:
Replaced by #155

PR #125, opened 2020-05-20
Add Swift 5.2 support

howlingblast:
Change Log

- Fix a crash due to order of conditions
- Change dependency of the Stencil package to the master branch; this is not optimal, but Stencil didn't have a release for a long time and Swift 5.2. support is only available on the master branch
- Adapt code to Swift 5.2/Stencil master (API changes)

### DIFF
--- a/.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
+++ b/.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Workspace
+   version = "1.0">
+   <FileRef
+      location = "self:">
+   </FileRef>
+</Workspace>
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:4.0
+// swift-tools-version:5.2
 import PackageDescription
 
 let package = Package(
@@ -7,7 +7,7 @@ let package = Package(
       .library(name: "StencilSwiftKit", targets: ["StencilSwiftKit"])
   ],
   dependencies: [
-      .package(url: "https://github.com/stencilproject/Stencil.git", .upToNextMinor(from: "0.13.0"))
+      .package(url: "https://github.com/stencilproject/Stencil.git", .branch("master"))
   ],
   targets: [
     .target(
@@ -22,6 +22,5 @@ let package = Package(
         "StencilSwiftKit"
       ]
     )
-  ],
-  swiftLanguageVersions: [4]
+  ]
 )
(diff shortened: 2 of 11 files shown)


Sources/StencilSwiftKit/Filters+Strings.swift
```diff
@@ -95,7 +95,7 @@ extension Filters.Strings {
     while idx < scalars.endIndex, let scalar = UnicodeScalar(scalars[idx].value), characterSet.contains(scalar) {
       idx = scalars.index(after: idx)
     }
-    if idx > scalars.index(after: start) && idx < scalars.endIndex,
+    if idx < scalars.endIndex, idx > scalars.index(after: start),
```
djbe, line 98: I'm assuming this fixes the crash?

Tests/StencilSwiftKitTests/CallNodeTests.swift
```diff
@@ -79,7 +79,7 @@ class CallNodeTests: XCTestCase {
   }
 
   func testRenderFail() {
-    let context = Context(dictionary: [:])
+    let context = Context(dictionary: ["": ""])
```
djbe, line 82: Why is this needed? (same for the lines below)

Package.swift
```diff
@@ -7,7 +7,7 @@ let package = Package(
       .library(name: "StencilSwiftKit", targets: ["StencilSwiftKit"])
   ],
   dependencies: [
-      .package(url: "https://github.com/stencilproject/Stencil.git", .upToNextMinor(from: "0.13.0"))
+      .package(url: "https://github.com/stencilproject/Stencil.git", .branch("master"))
```
djbe, line 10: I'd rather avoid this if possible, I can probably push through a Stencil release if needed.

Do we have Swift 5 support in Stencil yet? (it's been a while)

.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
djbe, line 1: Pretty sure we should add this folder to gitignore.

djbe (changes requested):
As mentioned in comments, It's a bad idea to depend on master, I'll see about making a Stencil release. Missing from this PR is an update to the podspec/Podfile.

AliSoftware:
PR #127 is pending and will soon update StencilSwiftKit to latest version of Stencil. This means we can stop pointing to the master branch in this PR, and that the changes like [this one](https://github.com/SwiftGen/StencilSwiftKit/pull/125#discussion_r428676384) will already be that other PR.

djbe:
Thank you for your contribution.

I know it's been a while, and in the mean time, this has been superseded by #127, which (amongst other things) adds support for Swift 5.

PR #118, opened 2019-03-02
CocoaPods 1.7.0 beta

djbe:
Similar to https://github.com/SwiftGen/SwiftGen/pull/619.

Note: based on #116, so this'll stay a draft PR until that's merged.

PR #127, opened 2020-08-24
Update to Stencil 0.14.0 (and Swift 5)

fortmarek:
Since Stencil has a new release, it'd be great to update it for `StencilSwiftKit`, too!

Apart from updating the version in `Package.swift`, some minor changes have been necessary


fortmarek:
The build is failing due to `Error untarring cache: Error extracting tarball` - I can't find what the problem could be and have zero experience with CircleCI, would you have any insights maybe, @AliSoftware?

AliSoftware:
Mmmh interesting, I got a very similar error in CircleCI for a completely different project and repo at work. Maybe CircleCI is being unstable recently about that? 🤔

Another reason might be if CircleCI updated something inside the VM images (like the version of bundler or the absolute path where the project is cloned inside the VM, changing the path of the cache being restored) without telling us…?

Two things to try:
1. Simply re-trigger the CI build. Maybe that was a one-off bug of CircleCI itself
2. Force-invalidate the cache so that build and subsequent ones use a new cache. You can do that by updating the cache key used in `.circleci/config.yml

I'll take care of trying both options in a minute.

AliSoftware:
@fortmarek Although unlikely, this _might_ also be because you're using a fork and not the main repo.
Now that you're part of the Core Contributors to the SwiftGen project, you're supposed to be able to create branches directly on the main repo instead of having to use a fork. Can you confirm me that is the case? (at least for your future PRs in SwiftGen's repos)

fortmarek:
Thanks for looking into it so quickly @AliSoftware! The fork could be why this is happening, but it seems I have not been added to core contributors - at least I don't see myself being part of the organization and I can not create a new branch in the repo itself.

AliSoftware:
Ah, right, I've added you to the `stencilproject` org but not `SwiftGen`'s. This should be fixed now 😅

AliSoftware:
Ok, seems CI is working again.

Some jobs are still failing, but for good reason now (found 1 violation during lint, and compile error on usage of `.components`) that actually needs fixing in your PR.

Package.swift
```diff
@@ -4,10 +4,10 @@ import PackageDescription
 let package = Package(
   name: "StencilSwiftKit",
   products: [
-      .library(name: "StencilSwiftKit", targets: ["StencilSwiftKit"])
+    .library(name: "StencilSwiftKit", targets: ["StencilSwiftKit"])
   ],
   dependencies: [
-      .package(url: "https://github.com/stencilproject/Stencil.git", .upToNextMinor(from: "0.13.0"))
+    .package(url: "https://github.com/stencilproject/Stencil.git", .upToNextMajor(from: "0.14.0"))
```
AliSoftware, line 10: When updating the `Package.swift`, you need to also [update the dependency in the `StencilSwiftKit.podspec` too](https://github.com/SwiftGen/StencilSwiftKit/blob/stable/StencilSwiftKit.podspec#L30). Once you've updated the `.podspec`, don't forget to run `bundle exec pod install` to update the project's `Podfile.lock` accordingly and commit the change to that lockfile too.

That's very likely why the CI doesn't compile when trying to build with Xcode (as opposed to SPM), as the Xcode project is still pointing to 0.13.x.

_(one day we'll configure [Danger](https://danger.systems/ruby/) on this repo like we did in the main SwiftGen repo, so that it will automatically warn you about that…)_

AliSoftware:
@fortmarek If you need to update the Xcode version for your build to work, you'll also need to change line 12 of `.circleci/config.yml` to tell CircleCI to use a VM which contains that Xcode pre-installed. See [here](https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions) for the list of VM images provided by CircleCI and their corresponding installed Xcode versions.

StencilSwiftKit.xcodeproj/project.pbxproj
```diff
@@ -1,463 +1,1570 @@
...
+   archiveVersion = "1";
+   objectVersion = "46";
+   objects = {
+      "OBJ_1" = {
+         isa = "PBXProject";
+         attributes = {
+            LastSwiftMigration = "9999";
+            LastUpgradeCheck = "9999";
+         };
+         buildConfigurationList = "OBJ_2";
```
AliSoftware: Seems like a strange diff for a pbxproj, with the list of `objects` being of the form `OBJ_nn` instead of being UUIDs like usual, and with `LastSwiftMigration` set to `9999`, etc…

How was this xcodeproj generated/updated? (Is it twist-generated? If so, would be better to do such a change in a separate PR to keep the diff focused on just the update of Stencil to 0.14)

fortmarek: Yeah, I thought the `.xcodeproj` was being created with swift package manager - which turns out not to be the case 😄 I'll revert the change. Thanks for the pointers, I'll ping you once I feel this PR is ready to be reviewed/I get stuck again 🙂

fortmarek:
I think this is ready for another review, @AliSoftware. One thing that will need to be resolved at some point (but I would not do it in this PR) is that the currently-used docker image has ruby with version 2.3.1. But if you run `bundle update`, you'd get `cocoapods-downloader` with version `1.4.0` which requires ruby `>= 2.4.0`. Unfortunately, we'll have to spin up our own docker image or find a different alternative from the jazzy docker image to solve this.

Gemfile
```diff
@@ -10,8 +10,8 @@ end
 
 # In addition to :build, for contributing
 group :development do
-	gem 'cocoapods', '~> 1.6.0'
-	gem 'rubocop', '~> 0.65'
+	gem 'cocoapods', '~> 1.7.0'
```
AliSoftware: I'd actually suggest that we change that version constraint to `~> 1.8` or even `~> 1.9`

There's no reason to constraint it to a specific `major.minor.x` (and only allow the patch component to be bumped) as opposed to letting the minor version be bumped to. SemVer is all about the fact that there's no breaking change as long as you keep the same major version, and `~> x.y` is recommended over `~> x.y.z`. Not sure why we used 3 digits constraints in the past tbh.

Latest version of CocoaPods is 1.9.x, and has the huge advantage of being able to use the CDN instead of having to clone the whole master Spec repo (which takes much more disk space and is long to clone). The use of the spec CDN improves both disk space and performance of `pod install` and avoids the very long initial "Setting up CocoaPods master repo" step on initial runs. So we might as well use that latest version anyway.

AliSoftware (changes requested):
I see.

If you want to lock `cocoapods-downloader` to a lower version tho, the way to do it is not to modify the `Gemfile.lock` – you should never edit the lockfile manually, lockfiles are manifests generated by dependency manager as a result of the dependency tree resolution, not intended to be edited manually – but to force the version of that gem in the `Gemfile` instead.

Issue is, we should definitively update to a more recent version of CocoaPods than even 1.7, rather than using an older version of `cocoapods` and pinning `cocoapods-downloader` to an even earlier version (see review comment). So I think it would still be worth fixing, even if that means putting this PR on hold and fixing that in a separate PR then rebase this one on top of it.

Gemfile
fortmarek: I agree with the sem versioning, I blindly followed what was already used there.

Yeah, let's try to go update the Cocoapods version - the issue is that installing ruby `>= 2.4.0` is not so easy to do in Ubuntu as `2.3.x` is. I'll try to create a new docker image, but it's gonna take me some more time.

AliSoftware: Maybe you can try with one of the official Swift Linux images on [Docker Hub](https://hub.docker.com/_/swift), e.g. `swift:4.2`

It would also be worth updating StencilSwiftKit to Swift 5 first (obviously we should ultimately do that in a separate PR). I think that would incidentally solve a lot of the issues you're having here, because once we'll be updated to use Swift 5, we can point to a `swift:5.2` image in Docker, and it being a more recent image, we'll also have a very good chance that the ruby version would be more recent as well and compatible with what `cocoapods-downloader` expects

AliSoftware: I actually just did a very quick "Edit > Convert to Current Swift Syntax" in the Xcode project to convert to Swift 5 and apparently, there's not a single change in the source code needed to make it compile with Swift 5. So just changing `s.swift_version` in the `podspec` and re-doing a `bundle exec pod install` so that the corresponding target in Xcode has an updated `SWIFT_VERSION` should probably be enough.

Since we're planning to upgrade to CocoaPods 1.9, we might also as well:
 - Update the `s.cocoapods_version = '>= 1.4.0'` in the `podspec` to a more recent requirement (at least 1.8 if not 1.9)
 - Use the plural form of the parameter `s.swift_versions = ["4.2", "5.0"]` instead of the singular form to indicate that we support both 4.2 and 5.0?

fortmarek: Unfortunately, the official Swift Linux image does not have `bundle` installed. What I can do is create a new docker image that will use Swift's image as a base and then installs ruby with the version we need + `bundle`. I tried to find already-existing suitable image, but could not find any.

djbe:
Note that I made PRs for Cocoapods 1.7 and Swift 5 quite a while ago: #118 and #119 respectively. There may be some bits in there that may be usable here if needed (haven't checked yet).

Gemfile
djbe: @AliSoftware I'd recommend setting `s.cocoapods_version` to the lowest one allowed (only needed for swift_versions property).

AliSoftware: Yeah, I've always had trouble navigating Docker Hub to find the image I want, especially knowing what's inside each (swift version AND ruby version)

I think before trying that, it'd be better to convert the project to Swift 5 first (see @djbe 's comment below, there is an open PR already pending that might need refreshing) and retry. Once we use Swift 5 and point to a Docker Image with Swift 5 (like `norionomura/jazzy:swift-5.2.0`) it might actually also already contain the more recent Ruby version we want…

fortmarek: I have tried using `norionmure/jazzy:swift-5.2.0` but that still contains ruby 2.3.x. They are even constraining the `cocoapods-downloader` [here](https://github.com/norio-nomura/docker-jazzy/blob/master/swift-5.2.5/Dockerfile#L19). The issue is that `apt-get install -y ruby-dev` does not let you install anything newer than `2.3.x`. 

I'll take a peek to that PR for the Swift version bump.

AliSoftware: Ah. 😐 

We have a `.ruby-version` file at the root of our repository, that is used when you use `chruby` to switch/install alternative Ruby versions. I'd suggest you bump the version of ruby in that file and try to see if we can install then use `chruby` in Norio's Docker image to install the ruby version we want…

fortmarek:
The CI seems to be green now. I have added Swift 5.2 image to a new repo [here](https://github.com/fortmarek/docker-StencilSwiftKit). I can transfer it to `SwiftGen` organization, but the problem is that the docker image will still reference my docker account (and for that we'd need SwiftGen Docker ID which might not be worth the hassle ?). I was also not able to configure with `chruby` (I tried hard, but ultimately failed ... but feel free to try it out by yourself). The current ruby version it uses is 2.5.1, which is what I put in `.ruby-version`.

I think it'd also be great to have only one `Package.swift` as it's easy to forget about the other, so how much are we keen on supporting Swift 4.0? Right now, this PR drops the support for it and keeps it only for Swift >= 4.2, but I can change that.

Apart from that, I have tried to update the Swift and Pods version where I deemed appropriate, but it definitely deserves another look.

AliSoftware:
I won't have time to review the PR today but just to answer your questions:

- I think we can keep pointing to your Docker image for now; I agree it's not worth the hassle to transfer it. We can always reevaluate later.
- I don't know much about Docker tbh but I was surprised when looking at your Dockerfile to see mention of ruby-0.7.1; didn't we want 2.4?!

fortmarek:
> I don't know much about Docker tbh but I was surprised when looking at your Dockerfile to see mention of ruby-0.7.1; didn't we want 2.4?!

That was version of `ruby-install`, but come to think of it, it was not necessary anymore anyway; the current ruby version is installed with `rubygems`, so I simplified the current `Dockerfile` 👍

fortmarek:
I think this is ready for another review @AliSoftware 🙂

Package.swift
```diff
@@ -23,5 +23,5 @@ let package = Package(
       ]
     )
   ],
-  swiftLanguageVersions: [4]
+  swiftLanguageVersions: [.v4_2, .version("5")]
```
AliSoftware, line 26: Shouldn't we migrate to `// swift-tools-version:5.0` and thus be able to use `.v5` here?

Sources/StencilSwiftKit/Environment.swift
```diff
@@ -7,7 +7,7 @@
 import Stencil
 
 public extension Extension {
-  public func registerStencilSwiftExtensions() {
+  func registerStencilSwiftExtensions() {
```
AliSoftware, line 10: Why this change?

StencilSwiftKit.xcodeproj/project.pbxproj
```diff
@@ -205,6 +211,7 @@
 			developmentRegion = English;
 			hasScannedForEncodings = 0;
 			knownRegions = (
+				English,
 				en,
```
AliSoftware, line 208: We should probably remove `en` if the new official value to specify that locale is supposed to be `English`. Or should that be the other way around?

StencilSwiftKit.xcodeproj/project.pbxproj
```diff
@@ -357,7 +364,7 @@
 				ONLY_ACTIVE_ARCH = YES;
 				SDKROOT = macosx;
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
-				SWIFT_VERSION = 2.3;
+				SWIFT_VERSION = 5.0;
```
AliSoftware, line 368: 😱 what a big leap 😂

.circleci/config.yml
```diff
@@ -6,37 +9,20 @@ defaults:
...
+      xcode: "11.4.1"
     shell: /bin/bash --login -eo pipefail
   - &linux-config
     environment:
       CIRCLE_ARTIFACTS: /tmp/circleci-artifacts
       CIRCLE_TEST_REPORTS: /tmp/circleci-test-results
       BUNDLE_PATH: vendor/bundle
     docker:
-      - image: norionomura/jazzy:swift-4.2.0
+      - image: marekfort/docker-stencilswiftkit:swift-5.2
```
AliSoftware: The only nit I'd make here is that I think it's a missed opportunity to have a more generic Docker image name — after all this one could be used by any other project that might require just the same ruby version, homebrew, and similar tools… but 🤷 that's not a big deal and could be kept for a later migration if we really want to.

AliSoftware (commented):
Just small nits and questions (but looks good overall!)

Sources/StencilSwiftKit/Environment.swift
schemers: most likely due to getting this compiler error if you try and mark the `func` as `public` when the extension is marked as `public`:

```
'public' modifier is redundant for instance method declared in a public extension
```

AliSoftware: Oh, good point! 👍

AliSoftware:
Just received an email from CircleCI about docker rate-limiting coming soon, and was wondering if there's any risk affecting us with our use of Docker in this PR…

> Hi there, 
> 
> On November 1st, [Docker Hub](https://go.circleci.com/F6WMG00hB0Z0u2uK000H2k4) will begin [limiting](https://go.circleci.com/gi0W4kG0K206MH02Z000Cuu) anonymous image pulls. We want to make sure you know how you might be impacted and what you can do to avoid interruptions to your workflow. 
> 
> Adding [Docker authentication](https://go.circleci.com/ek0WjH02000MZu0K46DGu02) to your pipeline config is the easiest way to avoid any service disruptions. If you use the [Docker executor](https://go.circleci.com/m000kE2040K26GuWMuk0ZH0) or pull Docker images when using the [machine executor](https://go.circleci.com/gl0W4kG0K206MH02Z000Fuu) on CircleCI, we encourage you to authenticate. Because the anonymous API rate limits are based on IP addresses, they will impact CircleCI cloud customers. Authenticated users get higher per-user rate limits, regardless of IP. 
> 
> We are currently working on a partnership with Docker to minimize the impact of this change for our users and will share more details as we get them. 
> 
> For more information or to leave a question for us, please head over to [Discuss](https://go.circleci.com/LK42HM0060mu0ZGWu0G020k) or [contact support](https://go.circleci.com/t2Hu2K0uZW0GH00kM40006n). 
> 
> Thanks and happy building, 
> 
> - The CircleCI team

@fortmarek you seem to be more versed into Docker than I am, any idea of the potential impact for us here?

fortmarek:
I believe the new Docker restrictions are OK as they limit the number of free anonymous pulls - but that's something we should not be doing very often, so I think we're fine

djbe:
~Why are we using these custom docker images instead of the official Swift images? Missing software, such as ruby/gems, or some other reason?~

Never mind, just found your dockerfile:
https://github.com/fortmarek/docker-StencilSwiftKit/blob/main/swift-5.2/Dockerfile

I've made a repo to manage our docker image, and automate the build&push process to docker: https://github.com/SwiftGen/swift-docker. It's slightly based on @fortmarek with some other things I've found. In the process, I created a SwiftGen docker organisation, so now we can have docker access tokens and never worry about usage limits.

.circleci/config.yml
djbe: Tadaaa: https://hub.docker.com/repository/docker/swiftgen/swift

AliSoftware:
> Why are we using these custom docker images instead of the official Swift images? Missing software, such as ruby/gems, or some other reason?

That's the reason indeed. See [this comment thread above](https://github.com/SwiftGen/StencilSwiftKit/pull/127#discussion_r479658886) for more context 😉 

> I've made a repo to manage our docker image, and automate the build&push process to docker: https://github.com/SwiftGen/swift-docker. It's slightly based on @fortmarek with some other things I've found. In the process, I created a SwiftGen docker organisation, so now we can have docker access tokens and never worry about usage limits.

Great!

Just one thing to be aware of tho, given the authentication phase you added in our CI config uses secret env vars, that means that the CI won't be able to run and authenticate on forks (on which CI secrets are not available, for security reasons), right? Not sure if that was already the case before or not. but worth to keep in mind

djbe:
Yeah the secrets aren't shared to forks, but shouldn't matter for such an "internal" repository. (talking about the docker repo)

The access tokens in CircleCI are set at a "team" level. I'm not 100% sure how that works with forks, this is what the docs say:
> To use environment variables set on the Contexts page, the person running the workflow must be a member of the organization for which the context is set.

So seems like it'll  only be used when we trigger the builds? Not that it really matters anyway, like @fortmarek mentioned, we wont be triggering too many builds on this repository.

AliSoftware:
Ok. sgtm!

Package.swift
AliSoftware: @djbe why did you mark this as resolved? Forgot the rationale for not using `// swift-tools-version:5.0` given we're now requiring Swift 5 to build anyway? Or do we still want to be Swift-4.2-compatible (tho if so, we don't check about this on CI, so not sure if we can guarantee such support officially 🤔 )

djbe: Note the other `Package@swift5.swift` file.

djbe: Good remark about testing swift 4.2, lemme see if I can create an image for it.

AliSoftware: Ah right.

I think we could consider merging this PR, and open a separate one to:

- Either drop Swift 4.2 support (removing `Package.swift` and renaming `Package@swift5.swift` to `Package.swift`)
- or keep supporting it by adding a Docker image for it and the relevant entry in CI config

djbe: Sounds good 👍

AliSoftware (approved):

fortmarek:
Thanks for taking this over the finish line @djbe! I've been swarmed with other stuff and couldn't get to it. Creating a new docker repo inside `Swiftgen` repository was something we were pondering but back then it did not seem to be worth it - which has changed with new restrictions, so well done 👏 Again, thanks for doing this 🙂

PR #129, opened 2020-10-09
Switch to Github Actions

djbe:
I think it might be a good idea to switch to Github Actions:
- Direct integration with PRs
- Seems to be faster from some quick testing. Especially any actions we run on `ubuntu-latest`, those are lightning fast.
- Swiftlint warnings now generate annotations directly in Github. See the example from the `test` commit, under "files changed".

The only thing that I'd consider a downside is that there is no SSH access to debug potential runner issues. But we've rarely (if ever) needed that feature.

Note: I have a branch on SwiftGen ready that does the same for that repo (includes danger).

-------

Ideas for a potential follow-up PR:
- Adding Danger to this repo
- Annotations for other files, such as changelog.


.github/workflows/cocoapods.yml
```diff
@@ -0,0 +1,29 @@
...
+      - '*.podspec'
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - 
+        name: Checkout
+        uses: actions/checkout@v2
```
AliSoftware, line 21: nit: why the odd format of leaving the first line blank?

.github/workflows/cocoapods.yml
```diff
@@ -0,0 +1,29 @@
...
+      - '.github/workflows/cocoapods.yml'
+      - '*.podspec'
+  pull_request:
+    paths:
+      - '.github/workflows/cocoapods.yml'
+      - '*.podspec'
+
+jobs:
+  lint:
+    name: Lint
```
AliSoftware:
```suggestion
    name: Pod Lint
```

.github/workflows/swiftlint.yml
```diff
@@ -0,0 +1,42 @@
...
+    branches: stable
+    paths:
+      - '.github/workflows/swiftlint.yml'
+      - '.swiftlint.yml'
+      - '**/*.swift'
+  pull_request:
+    paths:
+      - '.github/workflows/swiftlint.yml'
+      - '.swiftlint.yml'
+      - '**/*.swift'
```
AliSoftware, line 14:
```suggestion
    paths: &trigger_paths
      - '.github/workflows/swiftlint.yml'
      - '.swiftlint.yml'
      - '**/*.swift'
  pull_request:
    paths: *trigger_paths
```

.github/workflows/swiftlint.yml
```diff
@@ -0,0 +1,42 @@
...
+    runs-on: macos-latest
+    steps:
+      - 
+        name: Checkout
+        uses: actions/checkout@v2
+      -
+        name: Xcode Select
+        uses: devbotsxyz/xcode-select@v1.0.0
+        with:
+          version: ${{ env.xcode-version }}
```
AliSoftware: I haven't double-checked our Rakefile but do we actually need that, as opposed to our Rakefile prepending something like `DEVELOPER_DIR=$(find 12.0.0 version of Xcode via mdfind)`?

.github/workflows/cocoapods.yml
```diff
@@ -0,0 +1,29 @@
...
+on:
+  push:
+    branches: stable
+    paths:
+      - '.github/workflows/cocoapods.yml'
+      - '*.podspec'
+  pull_request:
+    paths:
+      - '.github/workflows/cocoapods.yml'
+      - '*.podspec'
```
AliSoftware, line 12:
```suggestion
    paths: &trigger_paths
      - '.github/workflows/cocoapods.yml'
      - '*.podspec'
  pull_request:
    paths: *trigger_paths
```

.github/workflows/swiftlint.yml
```diff
@@ -0,0 +1,42 @@
...
+      - '.github/workflows/swiftlint.yml'
+      - '.swiftlint.yml'
+      - '**/*.swift'
+
+env:
+  xcode-version: 12.0.0
+
+jobs:
+  lint:
+    name: Lint
```
AliSoftware:
```suggestion
    name: SwiftLint
```

.github/workflows/test-spm.yml
```diff
@@ -0,0 +1,78 @@
...
+on:
+  push:
+    branches: stable
+    paths:
+      - '.github/workflows/test-spm.yml'
+      - '**/*.swift'
+  pull_request:
+    paths:
+      - '.github/workflows/test-spm.yml'
+      - '**/*.swift'
```
AliSoftware, line 12:
```suggestion
    paths: &trigger_paths
      - '.github/workflows/test-spm.yml'
      - '**/*.swift'
  pull_request:
    paths: *trigger_paths
```

.github/workflows/test-spm.yml
```diff
@@ -0,0 +1,78 @@
...
+        name: Cache gems
+        uses: actions/cache@v2
+        with:
+          path: vendor/bundle
+          key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-gems-
+      - 
+        name: Cache SPM
+        uses: actions/Cache@v2
```
AliSoftware: Does it matter that the name of the action we use here is titlecased `Cache` while on line 28 it's all lowercase? Not sure which one is supposed to be the official one but would be nice to be consistent

.github/workflows/test-spm.yml
```diff
@@ -0,0 +1,78 @@
...
+        name: Checkout
+        uses: actions/checkout@v2
+      - 
+        name: Cache gems
+        uses: actions/cache@v2
+        with:
+          path: vendor/bundle
+          key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-gems-
```
AliSoftware, line 32: Any reason why we use this cache logic here, but on other workflows we just use `ruby/setup-ruby@v1` with `bundler-cache: true`? Why the diff?

.github/workflows/test-xcode.yml
```diff
@@ -0,0 +1,37 @@
...
+on:
+  push:
+    branches: stable
+    paths:
+      - '.github/workflows/test-xcode.yml'
+      - '**/*.swift'
+  pull_request:
+    paths:
+      - '.github/workflows/test-xcode.yml'
+      - '**/*.swift'
```
AliSoftware, line 12:
```suggestion
    paths: &trigger_paths
      - '.github/workflows/test-xcode.yml'
      - '**/*.swift'
  pull_request:
    paths: *trigger_paths
```

.github/workflows/test-xcode.yml
```diff
@@ -0,0 +1,37 @@
...
+    runs-on: macos-latest
+    steps:
+      - 
+        name: Checkout
+        uses: actions/checkout@v2
+      -
+        name: Xcode Select
+        uses: devbotsxyz/xcode-select@v1.0.0
+        with:
+          version: ${{ env.xcode-version }}
```
AliSoftware: Same q as before, shouldn't our Rakefile take care of the auto-select?

Scripts/SwiftLint.sh
```diff
@@ -3,23 +3,28 @@
 PROJECT_DIR="${PROJECT_DIR:-`cd "$(dirname $0)/..";pwd`}"
 SWIFTLINT="${PROJECT_DIR}/Pods/SwiftLint/swiftlint"
 CONFIG="${PROJECT_DIR}/.swiftlint.yml"
+if [ $CI ]; then
+	REPORTER="--reporter github-actions-logging"
```
AliSoftware, line 7: 😍

rakelib/spm.rake
```diff
@@ -6,13 +6,13 @@ if File.file?('Package.swift')
...
       Utils.print_header 'Compile using SPM'
-      Utils.run('swift build', task, xcrun: true)
+      Utils.run('swift build --enable-test-discovery', task, xcrun: true)
     end
 
     desc 'Run SPM Unit Tests'
     task :test => :build do |task|
       Utils.print_header 'Run the unit tests using SPM'
-      Utils.run('swift test', task, xcrun: true)
+      Utils.run('swift test --enable-test-discovery', task, xcrun: true)
```
AliSoftware, line 15: 🤯

rakelib/utils.rake
```diff
@@ -97,14 +97,12 @@ class Utils
 
   # run a command, pipe output through 'xcpretty' and store the output in CI artifacts
   def self.xcpretty(cmd, task, subtask)
-    name = (task.name + (subtask.empty? ? '' : "_#{subtask}")).gsub(/[:-]/, '_')
     command = [*cmd].join(' && ')
 
     if ENV['CI']
-      Rake.sh "set -o pipefail && (#{command}) | tee \"#{ENV['CIRCLE_ARTIFACTS']}/#{name}_raw.log\" | " \
-        "bundle exec xcpretty --color --report junit --output \"#{ENV['CIRCLE_TEST_REPORTS']}/xcode/#{name}.xml\""
+      Rake.sh %(set -o pipefail && (#{command}) | bundle exec xcpretty --color --report junit)
```
AliSoftware, line 103: I'm quite new to GHA, how are raw logs and artefacts handled? Does that change mean we won't be able to access the raw logs from the GitHub Action UI?

AliSoftware (commented):
Couple of additional questions:

 - How are artifacts handled in GitHub Actions? I saw you removed the raw logs from being created and didn't port the logic from CircleCI workflows that managed storing output logs into specific folders then save those folders as artefacts, does GitHub magically do that for us transparently?
 - Cache: why do we have some inconsistency between workflow (some doing explicit cache, some seems to do it implicitly via an external action); and does GHA auto-store cache at end of process as soon as there's a restore step? (as opposed to CircleCI which explicitly requires one step to restore previous cache at beginning of job, and one step to store new cache at the end)

.github/workflows/cocoapods.yml
djbe: Dunno, saw this format in some examples and other docs, and looks a bit clearer to me in editor 🤷

.github/workflows/swiftlint.yml
djbe: Github does not support this yaml ~thing~ anchors/aliases **yet** (https://github.community/t/support-for-yaml-anchors/16128/43)

.github/workflows/swiftlint.yml
djbe: May be a leftover from when I had this action directly invoke the SwiftLint script (sh), will have to check.

.github/workflows/cocoapods.yml
djbe: Same as above, github actions don't support fancy schmancy yaml yet.

.github/workflows/test-spm.yml
djbe: Because unfortunately the `ruby/setup-ruby` step/action only works on the default ubuntu images GitHub provides (and it is extremely fast on those). 

This job runs on our custom swift docker images, which don't have that, and the action is broken there.

.github/workflows/test-spm.yml
djbe: Good catch, didn't notice that (copy paste)

rakelib/utils.rake
djbe: Few things:
- We actually never really used those artifacts? Point of them was to be able to maybe analyse them in some follow-up step
- You can simply see **all** the output of a job/step right here in GitHub, there's no extra output

.github/workflows/swiftlint.yml
AliSoftware: Dammit

.github/workflows/test-spm.yml
AliSoftware: Gotcha. Might be worth a YAML comment to not wonder the same thing again later.
([edit] just saw you already did 👍)

.github/workflows/cocoapods.yml
AliSoftware: Fair enough

.github/workflows/test-spm.yml
```diff
@@ -0,0 +1,72 @@
...
+    name: Linux
+    runs-on: ubuntu-latest
+    container: swiftgen/swift:5.3
+    steps:
+      - 
+        name: Checkout
+        uses: actions/checkout@v2
+      - 
+        # Note: we can't use `ruby/setup-ruby` on custom docker images, so we
+        # have to do our own caching
```
AliSoftware, line 25: 👍

AliSoftware: Maybe worth exploring in the future to how to create our own reusable GitHub actions but compatible with custom docker images? But no need to do that now, just a thought to keep in mind for one day

djbe: Here's one issue I found BTW, which gives some extra info. GitHub does a bunch of custom stuff in their ubuntu images which we don't have access to. If at some point they release some sort of Dockerfile, we can maybe base our images on theirs.

https://github.com/ruby/setup-ruby/issues/88

rakelib/utils.rake
AliSoftware: There's a difference between the output and the artefacts though. The output is `xcpretty`-fied, while the point of having the artefacts in CircleCI is because – thanks to `tee` being inserted before `xcpretty` – the files we stored in the artefacts contained the raw logs (much more verbose) as opposed to the short/prettified ones that were printed on stdout. And I just checked and GitHub Action seems to support artefacts as well (actually when you click on a check on a pull request to see the details you can then click again on a specific job run and see its artefacts if it had any).

Tbh the point of those raw outputs is that if there's something wrong, sometimes the prettified output is not verbose enough to understand the real underlying raw error spit out by Xcode, and being able to look at the raw logs could help.  But in practice indeed I don't think we used those raw logs at any point – especially for SSK – so probably not reproducing the behavior there anyway at least for now.

AliSoftware (approved):

rakelib/utils.rake
djbe: We can easily add them back at some (later) point if we really want/need them 🤷

AliSoftware: Exactly

PR #130, opened 2020-10-14
Convert project to SPM

djbe:
Similar to https://github.com/SwiftGen/SwiftGen/pull/793, but much simpler in scope as we already have full SPM support in this project.

This just:
- Removes the old Xcode projects
- Removes CocoaPods (as a dependency/tool)
- Use commit hooks (Komondor) to run SwiftLint
- Adapts Rake and CI as needed


djbe:
Allright so this is slightly weird, and I'm not sure how we should deal with it:
- With #129 merged, we now run certain jobs only if specific files have changed
- I updated the list of required checks before a merge to those from #129 
- But this PR does not change the `podspec` file, so doesn't trigger a `Pod Lint` check...

Do we remove that pod check from the list of required steps? What about PRs where there's no code change, such as a doc PR? Should we remove all checks from being required?

PR #160, opened 2022-07-29
Add `swiftIdentifier` mode `valid`

### LINKED ISSUE #101: SwiftIdentifier filter: add modes for more flexibility
The current modes will always uppercase the first letter of each component, which is detrimental for some users that prefer to have the same keys as (for example) in their strings files.

Our current modes are:
- `normal`: converts to a valid identifier, uppercases first letter of each component
- `pretty`: same as normal + `snakeToCamelCase`

I propose adding some sort of `raw` mode, that only converts to a valid identifier, leaving the rest of the string as is. If need be, we can consider renaming the current modes to new names that better reflect their functionality.

For backwards compatibility, if we rename modes, we can always keep the old names as aliases for a few versions.

djbe:
Adds the mode discussed in #101 for minimal modification of values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@
 * Added an `import` tag for reusing macro's in multiple templates from a common imported file.  
   [David Jennes](https://github.com/djbe)
   [#111](https://github.com/SwiftGen/StencilSwiftKit/pull/111)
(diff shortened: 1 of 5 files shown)


SwiftGen-Eve:
Hey 👋 I'm Eve, the friendly bot watching over SwiftGen 🤖

Thanks a lot for your contribution!



---

Seems like everything is in order 👍 You did a good job here! 🤝

<p align="right" data-meta="generated_by_danger">
  Generated by :no_entry_sign: <a href="https://danger.systems/">Danger</a>
</p>

PR #162, opened 2022-08-02
Fix version for Komondor

redryerye:
[Komondor](https://github.com/shibapm/Komondor) had a release for [v.1.1.4](https://github.com/shibapm/Komondor/compare/1.1.3...1.1.4) where it added unnecessary dependencies that will only be used in Komondor's development process.
It would be great if we could minimize dependencies to avoid SPM conflicts by not updating Komondor to v1.1.4.

Right now, Sourcery as a SPM plugin is [facing a conflict](https://github.com/krzysztofzablocki/Sourcery/issues/1023#issuecomment-1180284670) because of this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,9 @@ _None_
 
 ### Internal Changes
 
-_None_
+* Pin Komondor to 1.1.3, to avoid issues with SPM for other packages depending on this package.  
+  [redryerye](https://github.com/redryerye)
+  [#162](https://github.com/SwiftGen/StencilSwiftKit/pull/162)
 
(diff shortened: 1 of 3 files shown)


krzysztofzablocki:
@djbe this is a big issue for Sourcery

djbe:
Ugh this is annoying TBH, Komondor should release a new version with those dependencies hidden again 🤦 

But then again, should we be hiding dependencies like Komondor ourselves as well? What about testing dependencies like Spectre (from Stencil)? Although Komondor is definitely the biggest issue, because it includes quite a lot of big packages.

djbe (approved):

